PR #70, opened 2016-09-06
when you run get-reviewers.py without any arguments, it gets all revi…

mspector:
…ewers for all notes


mspector:
I know I'm not supposed to merge my own PRs, but I wanted to get this out ASAP - it's a small change that we can revert if needed

PR #78, opened 2016-09-11
updated assign-reviewer

mspector:


iclr2017/python/assign-reviewer.py
```diff
@@ -78,10 +78,16 @@ def get_reviewer_group(openreview, reviewer, paper_number, conflict_list):
         if str(c.strip()):
             conflict_list.append(str(c.strip()))
 user_conflict = None
+conflicted_users = []
+
 for c in conflict_list:
     group = openreview.get_group(c)
     if group.members and openreview.user['id'] in group.members:
         user_conflict = c    
+        conflicted_users.append(openreview.user['id'])
```
melisabok: conflicted user can be or the user than run the script or the reviewer. I would display a different error message when the conflict is with the user or with the reviewer.

iclr2017/python/assign-reviewer.py
```diff
@@ -78,10 +78,16 @@ def get_reviewer_group(openreview, reviewer, paper_number, conflict_list):
...
+    if group.members and reviewer in group.members:
```
melisabok: I'm not sure if this is the right conflict validation. We would need to get all the transitive members of the user|reviewer and check if one of them are in the group.members.

PR #86, opened 2016-09-16
Feature/remove edges collections

melisabok:


mspector:
this all looks good. I like the use of the ".template" file

PR #91, opened 2016-09-21
Fix/reviewer assignment bug

mspector:


melisabok:
You only removed the file, do you need to modify the template?

mspector:
No, the template was already correct, I just hadn't generated a new js file from it

PR #99, opened 2016-09-22
added export-members.py

mspector:


iclr2017/python/export-members.py
```diff
@@ -0,0 +1,39 @@
...
+args = parser.parse_args()
+
+## Initialize the client library with username and password
+if args.username!=None and args.password!=None:
+    openreview = Client(baseurl=args.baseurl, username=args.username, password=args.password)
+else:
+    openreview = Client(baseurl=args.baseurl)
+
+
+group = openreview.get_group(args.group)
```
melisabok: Please check if the group exists

iclr2017/python/export-members.py
```diff
@@ -0,0 +1,39 @@
...
+members = group.members
+if args.sort!=None and args.sort.lower()=='true':
+	members = sorted(members, key=lambda s: s.lower())
+with open(args.output, 'wb') as outfile:
+    csvwriter = csv.writer(outfile, delimiter=',')
+
+    for m in members:
```
melisabok: couldn't you send the members directly?

mspector: It seems like this would work, but I changed it to something cleaner anyway:

http://stackoverflow.com/questions/22999216/writing-a-list-of-sentences-to-a-single-column-in-csv-with-python

PR #100, opened 2016-09-22
Fix/add email groups

mspector:


iclr2017/python/email-group[template].py
```diff
@@ -0,0 +1,62 @@
...
+#################################################
+
+recipients = args.group if args.group!=None else ""
+
+groupToMail = openreview.get_group(recipients)
+
+if type(groupToMail)==Group:
+    for member in groupToMail.members:
+        print "Sending message to "+member
+        openreview.send_mail(subjectline, [member], message)
```
Comment (melisabok) [inline iclr2017/python/email-group[template].py]: The REST Api can receive a list of groups as recipients. And for each group tries to get all the members of that groups that are emailable. 
So if you want to send an email to all the Pcs, you just need to send the group ICLR.cc/2017/pcs

PR #114, opened 2016-10-11
added dblp directory and tutorial section

mspector:


dblp/python/superuser-init.py
```diff
@@ -0,0 +1,127 @@
...
+                'value-regex': '[\\S\\s]{0,5000}',
+                'required':False
+            },
+            'authors': {
+                'description': 'Comma separated list of author names, as they appear in the paper.',
+                'order': 3,
+                'value-regex': '[^,\\n]*(,[^,\\n]+)*',
+                'required':False
+            },
+            'author_emails': {
```
melisabok: We should change this field to authorids

dblp/python/superuser-init.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+    reply = {
+        'forum': None,
+        'replyto': None,
+        'readers': {
+            'description': 'The users who will be allowed to read the above content.',
+            'values': ['everyone']
+        },
+        'signatures': {
```
melisabok: signatures and writes will be DBLP.org/upload, I think we should create this group too,

andrewmccallum:
I think there should be both "authors" and "authorids"
The first are their string names, exactly as they appear in the DBLP BibTeX record.
The second will be email address, with "null" for authors for whom we don't have an address.

mspector:
yes, this authors and authorids are both included

PR #116, opened 2016-10-13
changed author_emails to authorids in iclr2017 scripts

mspector:


iclr2017/process/commentProcess.js
```diff
@@ -8,7 +8,7 @@ function(){
 
     var getAuthorEmails = function(origNote){
       console.log('get author emails initiated')
-      var origNoteAuthors = origNote.content.author_emails.trim().split(",");
+      var origNoteAuthors = origNote.content.authorids.trim().split(",");
```
melisabok: authorids is an array, right? the trim and split functions are not needed.

PR #120, opened 2016-10-26
Feature/matching assignment

agankur21:
Added scripts for 
1)Creating a new conference UAI2017 : in the folder uai2017/
2).Created 3 scripts in the uai2017/python
   - superuser-init.py
   -sample_notes_reviewer.py
   -sample_notes_paper.py
3).Created a folder dto/uai2017 for capturing all the objects and constants related to the scripts
   *This folder decouples the object definition from the script ,making the script agnostic of the object definitions


uai2017/python/superuser-init.py
```diff
@@ -0,0 +1,338 @@
...
+                'value-regex': '.{1,100}',
+                'required':True
+            },
+            'authors': {
+                'description': 'Comma separated list of author names, as they appear in the paper.',
+                'order': 2,
+                'value-regex': '[^,\\n]+(,[^,\\n]+)*',
+                'required':True
+            },
+            'author_emails': {
```
melisabok: author_emails were renamed to author_ids: https://github.com/iesl/openreview-scripts/blob/master/iclr2017/python/superuser-init.py#L221

author and author_ids are values-regex instead of value_regex. 

I think @agankur21 need to update the last changes of iclr2017, what do you think @mspector

mspector: Yes, I agree. The changes that need to be made are:

1. authors_emails --> authorids
2. change authors, authorids, and conflicts from "value-regex" to "values-regex". This change will cause these fields to be saved as arrays, instead of strings ( e.g "Michael Spector, Melisa Bok, Ankur Agarwal" --> ["Michael Spector","Melisa Bok","Ankur Agarwal"] )
3. the regex strings for the above fields should be changed so that they refer to each item in the array, instead of regex for the whole string. For now I think it's fine to just set these as /.*/ (match any character).

uai2017/process/commentProcess.js
```diff
@@ -0,0 +1,161 @@
...
+    console.log('comment process initiated')
+    var or3client = lib.or3client;
+    
+    var list = note.invitation.replace(/_/g,' ').split('/');
+    list.splice(list.indexOf('-',1));
+    var conference = list.join(' ')
+
+    var getAuthorEmails = function(origNote){
+      console.log('get author emails initiated')
+      var origNoteAuthors = origNote.content.author_emails.trim().split(",");
```
mspector: replace "author_emails.trim()" with "authorids" (without trimming) across all process functions.

you can see the most up-to-date version of the ICLR 2017 process functions to see this

agankur21:
I have updated this in the new commit

PR #140, opened 2016-11-10
superuser assign scripts

mspector:
assigns using the tilde group, if available. only runnable by the superuser

### DIFF
--- a/iclr2017/python/superuser-assign-reviewers.py
+++ b/iclr2017/python/superuser-assign-reviewers.py
@@ -1,0 +1,164 @@
+#!/usr/bin/python
+
+###############################################################################
+# Not to spec
+###############################################################################
+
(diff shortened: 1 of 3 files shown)


iclr2017/python/superuser-assign-reviewers.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if not '@' in areachair:
+            return False
+
+        return True
+    except IndexError:
+        return False
+
+def assign_reviewer(reviewer,paper_number):
+    notes = [note for note in submissions if str(note.number)==str(paper_number)]
```
melisabok: you could use the get notes by invitation and paper number. or save the submissions in a dictionary where the key is the paper number, so you don't need to iterate the collection for each assingments

PR #151, opened 2016-11-18
fixed get-areachairs error handling

mspector:


iclr2017/python/get-areachairs.py
```diff
@@ -34,23 +34,26 @@
...
         note = notes[0]
         message = []
         areachairs = openreview.get_group('ICLR.cc/2017/conference/paper'+str(note.number)+'/areachairs');
         for ac in areachairs.members:
             areachair_wrapper=openreview.get_group(ac)
             areachairNumber = ac.split('paper')[1].split('/areachair')[1]
             
-            pad = '{:40s}'.format("["+str(areachair_wrapper.members[0])+"] ")
-            message.append(pad+"areachair"+areachairNumber+" ("+ac+")")
+            if areachair_wrapper.members:
```
melisabok: you need to check the length of the collection

iclr2017/python/get-areachairs.py
```diff
@@ -34,23 +34,26 @@
 
 if args.paper_number!=None:
     
-    try:
+    if notes:
```
melisabok: you need to check the length of the collection

mspector: aren't I just checking to see if the list is empty? if so, this is the "proper" way to do it in python (according to my internet search)

http://stackoverflow.com/questions/53513/best-way-to-check-if-a-list-is-empty

PR #163, opened 2017-01-05
first version of get paper status for ICLR2017

pmandler:


iclr2017/python/get-all-paper-status.py
```diff
@@ -0,0 +1,159 @@
...
+    except ValueError:
+        print 'Type must be 1 (for text) or 2 (for csv).'
+        sys.exit()
+    if (output_type is not 1) and (output_type is not 2):
+        print 'Type must be 1 (for text) or 2 (for csv).'
+        sys.exit()
+  elif opt in ("-o", "--ofile"):
+    file_name = arg
+my_file = open(file_name, "w")
+
```
mspector: I think this is fine, but you may want to consider using the argparse library to do arguments instead. I think it's a bit easier. Check this script out for an example: [script](https://github.com/iesl/openreview-scripts/blob/master/iclr2017/python/export-members.py)

don't even worry about what the script does - just look at the first 28 lines to see how arguments get handled.

iclr2017/python/get-all-paper-status.py
```diff
@@ -0,0 +1,159 @@
...
+
+client = openreview.Client(username='OpenReview.net', password='OpenReview_beta', baseurl='https://openreview.net')
```
mspector: Whether you use argparse or not is up to you, but it is important that we not give away the superuser password.

In my previous comment, I shared with you an example [script](https://github.com/iesl/openreview-scripts/blob/master/iclr2017/python/export-members.py), which also demonstrates how you can use arguments to set the username, baseurl, and password.

Let me know if you need help doing this

pmandler: 1. I've added the arguments for baseurl, username and password but I don't think they work well.  When I tried --baseurl https://localhost:3000 it failed.  I'd like to test that for an unauthorized person it will give Anonymous instead of the actual reviewer names.
2. When I commit code it thinks my email is mandler@titian.cs.umass.edu instead of mandler@cs.umass.edu.  I don't

mspector: 1. the baseurl for localhost should be http://localhost:3000, instead of https (only the live site uses https)

2. there should be a setting somewhere in git that can help you change the default account name/email... I can help you out with this later today if you'd like

pmandler:
Good catch.  Now I’m getting the localhost, but it reports an error getting reviewer ids.  Is my database from a few weeks ago incompatible with the current code or is this just a test case that I need to handle better?

Thanks,
Pam
___________________
Pam Mandler
Center for Data Science <https://ds.cs.umass.edu/>
(413) 545-2561

mspector:
it would very much depend on what the error is. have you had a chance to investigate since then? we can look at it together after the UAI demo today, if you want

pmandler:
No I haven’t looked at it.  When is the UAI demo?  I would like to look at it with you because I’m not sure how to attack the problem.

Pam
___________________
Pam Mandler
Center for Data Science <https://ds.cs.umass.edu/>
(413) 545-2561

mspector:
It's at 3:00. I'll shoot you an email so we can take this off the comment thread

melisabok:
I think the changes are ok, @mspector ?

PR #166, opened 2017-01-09
added js dictionary file using templates

mspector:
note that spc_registrationProcess is now a .template as well


uai2017/process/uaidata.js
```diff
@@ -0,0 +1,95 @@
+{
```
melisabok: I like the idea!

What is this a js module with variables or a json object? I don't understand what it is. 

and why are they in all CAPS?

PR #178, opened 2017-01-11
Fix/wrap blinded authors

melisabok:


uai2017/python/superuser-init.py
```diff
@@ -41,6 +42,9 @@ def overwrite_allowed(groupid):
     else:
         return False
 
+call(["node", "../../scripts/processToFile.js", "../process/submissionProcess.template", "../process"])
+call(["node", "../../scripts/processToFile.js", "../process/spc_registrationProcess.template", "../process"])
+
```
mspector: nice! thank you!

PR #201, opened 2017-01-23
get the number of replies per paper per reviewer/areachair

melisabok:


iclr2017/python/get-stats.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/python
+
+"""
+Sends an email to the members of the group of your choice.
+
```
mspector: should change this description

melisabok: done

PR #203, opened 2017-01-24
fixed commentProcess function to filter out program chairs when they …

mspector:
…are the author of the paper

### DIFF
--- a/iclr2017/process/commentProcess.js
+++ b/iclr2017/process/commentProcess.js
@@ -110,7 +110,7 @@ function(){
       console.log('replytoNoteSignatures before filter: ' + replytoNoteSignatures);
 
       if (note.readers.indexOf('everyone') == -1){
-        replytoNoteSignatures=[];
+        replytoNoteSignatures = [];
       };
       console.log('replytoNoteSignatures after filter: '+replytoNoteSignatures);
       var comment_mail = {
--- a/iclr2017/process/commentProcess.js
+++ b/iclr2017/process/commentProcess.js
@@ -87,15 +87,20 @@ function(){
         //if any member of the PCs is an author, remove that PC from email recipients
         for (var i = 0; i < origNote.content.authorids.length; i++){
           var currentAuth = origNote.content.authorids[i];
+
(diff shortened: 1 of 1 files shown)


iclr2017/process/commentProcess.js
```diff
@@ -67,12 +67,36 @@ function(){
...
         if(signatureIdx>-1){
           pcs.splice(signatureIdx,1);
-        };
-        console.log('pcs after filter:',pcs);
+        }
+
+        //if any member of the PCs is an author, remove that PC from email recipients
+        for(var i=0; i<origNote.content.authorids.length; i++){
+          var currentAuth = origNote.content.authorids[i];
+          var pcIdx = pcs.indexOf(currentAuth);
```
melisabok: this only works when the email of the author is the same as the member of the pc group

iclr2017/process/commentProcess.js
```diff
@@ -67,12 +67,36 @@ function(){
       return or3client.or3request(or3client.grpUrl+'?id=ICLR.cc/2017/pcs',{},'GET',token)
       .then(result=>{
         var pcs = result.groups[0].members;
+
+        console.log('pcs before filter: '+pcs);
+
+        var workshopIdx = pcs.indexOf('ICLR.cc/2017/workshop');
+        if(workshopIdx>-1){
```
melisabok: Please, start to follow the js guidelines: http://standardjs.com/rules.html. Spaces after keywords.

iclr2017/process/commentProcess.js
mspector: yes, that's true. I could go through each author email and get groups by member, and filter the PCs that way. But that's a lot of calls. Do you think it's ok? or do you have another suggestion?

melisabok: Given that are so few pcs, I would do a check over the papers and be sure that emails are the same(authorids and pc members). Otherwise we can have a workaround in the process function.

mspector: Ok. if one of the PCs uses a different email address, should I change the authorid in the paper? or should I add the email address to the PC group?

PR #210, opened 2017-01-26
get acceptance status

pmandler:
This is an incremental change.  It has official reviews, meta-reviews and acceptance notes.  It does not have all a count of all interactions, and the 'names' are still says things like area chairs not actual names.


melisabok:
Great!!! thanks!

PR #211, opened 2017-01-27
Accept status with names

pmandler:
Add real names for reviewers and area chairs.
Include a line for AC's even if they haven't entered a meta review - this line looks like a meta-review with Recommendation of TBD.

### DIFF
--- a/iclr2017/python/get-acceptance-status.py
+++ b/iclr2017/python/get-acceptance-status.py
@@ -16,7 +16,7 @@ if args.username!=None and args.password!=None:
 else:
     client = openreview.Client(baseurl=args.baseurl)
 ## Initialize output file name
-file_name = "status.csv"
+file_name = "acceptances.csv"
 if args.ofile!=None:
     file_name = args.ofile
 
--- a/iclr2017/python/get-acceptance-status.py
+++ b/iclr2017/python/get-acceptance-status.py
@@ -1,0 +1,68 @@
+import argparse
+import openreview
+import csv
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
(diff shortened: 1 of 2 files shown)


iclr2017/python/get-acceptance-status.py
```diff
@@ -26,6 +27,12 @@
 acceptances = client.get_notes(invitation='ICLR.cc/2017/conference/-/paper.*/acceptance')
 reviews = client.get_notes(invitation='ICLR.cc/2017/conference/-/paper.*/official/review')
 
+headers = {'User-Agent': 'test-create-script', 'Content-Type': 'application/json',
+           'Authorization': 'Bearer ' + client.token}
+anon_reviewers = requests.get(client.baseurl + '/groups?id=ICLR.cc/2017/conference/paper.*/AnonReviewer.*',
+                              headers=headers)
+area_chairs = requests.get(client.baseurl+'/groups?id=ICLR.cc/2017/conference/paper.*/areachair1.*', headers = headers)
```
melisabok: there are a few cases where we have an areachair2 instead of areachair1. 

https://openreview.net/groups?id=ICLR.cc/2017/conference/paper.*/areachair2

So, this query is not going to give these groups.

iclr2017/python/get-acceptance-status.py
```diff
@@ -43,35 +50,80 @@ def get_score(note, content_type):
...
+            paper_status[paper_num][reviewer_id]['type'] = "meta-review"
+            paper_status[paper_num][reviewer_id]['recommendation'] = "TBD"
 
 # official reviews
 for note in reviews:
     paper_entry = get_paper_status_entry(note, '/official/review')
-    paper_entry['type'] = "review"
-    paper_entry['score'] = get_score(note, 'rating')
-    paper_entry['confidence'] = get_score(note, 'confidence')
+    if paper_entry is not None:
```
melisabok: I guess you can also write it like this:

```
if not paper_entry:
```

PR #234, opened 2017-02-14
first version get all notes associated with a submission

pmandler:


iclr2017/python/get-submission-notes.py
```diff
@@ -0,0 +1,82 @@
...
+    for paper_num in paper_status:
+        for note_id in paper_status[paper_num]:
+            if note_id != 'pdf':
+                row = []
+                row.append(paper_num)
+                row.append(paper_status[paper_num][note_id]['timestamp'])
+                row.append(paper_status[paper_num][note_id]['type'].encode('utf-8'))
+                row.append(paper_status[paper_num][note_id]['author'].encode('utf-8'))
+                row.append(paper_status[paper_num][note_id]['content'])
+                row.append(paper_status[paper_num]['pdf'].encode('utf-8'))
```
melisabok: the pdf url is an internal url, you can build the public url like this:

https://openreview.net/pdf?id=[note.id]

Example: https://openreview.net/pdf?id=BkbY4psgg

iclr2017/python/get-submission-notes.py
```diff
@@ -0,0 +1,82 @@
...
+    paper_status[paper.number]['pdf'] = paper.content['pdf']
+
+# add each note to the associated paper in paper_status
+for note in allnotes:
+    paper_num0 = note.invitation.split('paper')[1].split('/')[0]
+    note_type = note.invitation.split('paper' + str(paper_num0))[1]
+    # WHY does paper_num work and paper_num0 not?
+    paper_num = int(note.invitation.split('paper')[1].split(note_type)[0])
+    if paper_num in paper_status:
+        author = note.writers[0]
```
melisabok: It is better to use the first element of signatures instead of writers. There are some case they are not the same, and signatures is how the use signed the note.

melisabok:
Great, thanks!

PR #237, opened 2017-02-15
anonymize note authors

pmandler:
line 122 - there must be a better way to do this without looking through the entire dictionary - just where we are at and forward.

### DIFF
--- a/iclr2017/python/get-submission-notes.py
+++ b/iclr2017/python/get-submission-notes.py
@@ -107,7 +107,7 @@ for r in area_chairs.json():
 # make sure ALL names are in
 for paper_num in paper_status:
     for note_id in paper_status[paper_num]:
-        if note_id != 'pdf':
+        if note_id != 'paper_id':
             name = paper_status[paper_num][note_id]['author']
             if name not in reviewers:
                 reviewers[name] = name
@@ -138,12 +138,14 @@ with open(file_name, 'wb') as outfile:
     for paper_num in paper_status:
         for note_id in paper_status[paper_num]:
             if note_id != 'paper_id':
+                pdf_url = 'https://openreview.net/pdf?id='+paper_status[paper_num]['paper_id'].encode('utf-8')
+                author_name = 'reviewer'+str(reviewers[paper_status[paper_num][note_id]['author']])
                 row = []
                 row.append(paper_num)
                 row.append(paper_status[paper_num][note_id]['timestamp'])
                 row.append(paper_status[paper_num][note_id]['type'].encode('utf-8'))
-                row.append('reviewer'+reviewers[paper_status[paper_num][note_id]['author']])
+                row.append(author_name)
                 row.append(paper_status[paper_num][note_id]['content'])
-                row.append('https://openreview.net/pdf?id='+paper_status[paper_num]['paper_id'].encode('utf-8'))
+                row.append(pdf_url)
                 csvwriter.writerow(row)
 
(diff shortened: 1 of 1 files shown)


iclr2017/python/get-submission-notes.py
```diff
@@ -28,6 +29,12 @@
 submissions = client.get_notes(invitation='ICLR.cc/2017/conference/-/submission')
 allnotes = client.get_notes(invitation='ICLR.cc/2017/conference/-/paper.*/*')
 
+headers = {'User-Agent': 'test-create-script', 'Content-Type': 'application/json',
+           'Authorization': 'Bearer ' + client.token}
+anon_reviewers = requests.get(client.baseurl + '/groups?id=ICLR.cc/2017/conference/paper.*/AnonReviewer.*',
+                              headers=headers)
+area_chairs = requests.get(client.baseurl+'/groups?id=ICLR.cc/2017/conference/paper.*/areachair.*', headers = headers)
+
```
mspector: I should improve the python library to allow for this

iclr2017/python/get-submission-notes.py
```diff
@@ -52,9 +57,71 @@
...
+def check_profile(member):
+
+    if member.startswith('~'):
+        return member
+
+    if '@' in member:
+        if member in profile_by_email:
+            return profile_by_email[member]
+
+        response = requests.get(client.baseurl+'/user/profile?email=' + member, headers = headers)
```
mspector: you should be able to do client.get_profile(id_or_reviewer) here

iclr2017/python/get-submission-notes.py
```diff
@@ -52,9 +57,71 @@
...
+                reviewers[name] = name
+
+# now that we know the real ID of everyone, we hide all of the names....
+new_val = 0
+for r in reviewers:
+    name = reviewers[r]
+    if type(name) != int:
+        new_val += 1
+        # it would be nice to start from 'r' and move forward
+        for y in reviewers:
```
mspector: doing something like this is easier when the object you're iterating over is an ordered list, instead of a dictionary. You'd have to turn this into a list somehow to do what you're describing - whether that means "enumerating" the dict, or some other way of doing it...

(I think if it's running in a reasonable amount of time, it's fine to leave it as is)

pmandler:
What does client.get_profile() take?  How much of the code can I get rid of?  

Pam
___________________
Pam Mandler
Center for Data Science <https://ds.cs.umass.edu/>
(413) 545-2561

mspector:
I mentioned this in my comment above: it takes an ID or an email address. It returns a Note which is constructed from the profile variable you're setting on line 83. 

This means you should change line 84 from profile['id'] to profile.id, or profile.to_json()['id']

mspector:
Let me know if you want me to go over it with you. Sorry for the lack of documentation! I just tried swinging by your office and you weren't there, but let me know.

pmandler:
This didn’t work for me. Says "AttributeError: 'Client' object has no attribute 'get_profile’“.  It seems I don’t have the latest code.  I just did a pull request.  No change.  Then I restarted "NODE_ENV=development node app.js” and it barfed.  Do I need to run grunt?

Pam
___________________
Pam Mandler
Center for Data Science <https://ds.cs.umass.edu/>
(413) 545-2561

mspector:
You may need to update your openreviewpy installation. To do that, go to
your terminal and enter:

pip install --upgrade openreviewpy

You should get a message stating that your version has been updated to
0.5.32

I'll be in soon and I can help you if you need it

PR #241, opened 2017-02-17
created get workshop submissions script

mspector:


iclr2017/python/get-workshop-submissions.py
```diff
@@ -0,0 +1,57 @@
...
+
+            for note in notes:
+                row = []
+                row.append(note.id)
+                row.append(note.number)
+                row.append(note.tcdate)
+                row.append(note.content['title'].encode('UTF-8'))
+                row.append(note.content['TL;DR'].encode('UTF-8'))
+                row.append(note.content['abstract'].encode('UTF-8'))
+                row.append(', '.join(note.content['keywords']).encode('UTF-8'))
```
pmandler: Will this mess up the columns?

mspector: I tried it, and it looks like it doesn't

melisabok:
It would be nice to have the original link in the file. So they can see which papers were already submitted in the conference. To get that information we need to change the OpenReviewPy, right? @mspector

PR #243, opened 2017-02-17
added original field to the csv output

mspector:


mspector:
to make this work, you'll need to upgrade OpenReviewPy to version 0.5.34

iclr2017/python/get-workshop-submissions.py
```diff
@@ -36,13 +36,14 @@
...
             csvwriter = csv.writer(outfile, delimiter=',')
-            fieldnames = ['id', 'number', 'date', 'title', 'TL;DR', 'abstract','keywords','authors','authorids','conflicts']
+            fieldnames = ['id', 'number', 'original', 'date', 'title', 'TL;DR', 'abstract','keywords','authors','authorids','conflicts']
             csvwriter.writerow(fieldnames)
 
             for note in notes:
                 row = []
                 row.append(note.id)
                 row.append(note.number)
+                row.append('https://openreview.net/forum?id=%s' % note.original if note.original else None)
```
melisabok: I guess note.original returns None, you don't need the if, right?

mspector: if note.original is None, then the field looks like this:

https://openreview.net/forum?id=None

PR #244, opened 2017-02-17
Feature/add recommend reviewer invitation

melisabok:


uai2017/python/superuser-init.py
```diff
@@ -526,6 +526,44 @@ def overwrite_allowed(groupid):
 
     invitations.append(bid_tag_invitation)
 
+    recommendation_tag_invitation = openreview.Invitation('auai.org/UAI/2017',
+        'Recommend/Reviewer',
+        readers=['everyone'],
+        writers=['auai.org/UAI/2017'],
```
mspector: I would use the CONFERENCE variable from uaidata.py

PR #251, opened 2017-02-23
Fix/user submission different email

mspector:


uai2017/process/usersubmissionProcess.js
```diff
@@ -36,7 +36,7 @@ function() {
         id: groupId,
         signatures: [CONFERENCE],
         writers: [CONFERENCE],
-        members: note.content.authorids,
+        members: note.content.authorids.concat(note.signatures),
```
melisabok: can we ensure that are not repeated emails? probably use _.union?

mspector: well, the signature will be a tilde ID, so I'm not sure that would work. I think disambiguating between the email and the tilde ID would be too complicated for what we're getting out of it

melisabok: I mean instead of using concat, use union to remove duplicates if exist.

PR #260, opened 2017-02-28
matching demo 2

mspector:


uai2017/python/metadata-init.py
```diff
@@ -0,0 +1,118 @@
...
+paper_metadata_reply = {
+    'forum': None,
+    'replyto': None,
+    'readers': {
+        'description': 'The users who will be allowed to read the above content.',
+        'values': [COCHAIRS, ADMIN] #who should be allowed to read UAI submissions and when?
+    },
+    'signatures': {
+        'description': 'How your identity will be displayed with the above content.',
+        'values-regex': 'OpenReview.net'
```
melisabok: I guess we should use the UAI admin user in the future, right?

uai2017/python/metadata-papers.py
```diff
@@ -0,0 +1,141 @@
...
+bids_by_number = defaultdict(list)
+bids_by_id = defaultdict(list)
+
+for b in bids:
+    n = client.get_note(b.forum)
+    bids_by_number[n.number].append(b)
+    bids_by_id[n.forum].append(b)
+
+
+recs = client.get_tags(invitation='auai.org/UAI/2017/-/Recommend/Reviewer')
```
melisabok: I changed the invitation, now we have one invitation per paper:

auai.org/UAI/2017/-/Paper.*/Recommend/Reviewer

PR #265, opened 2017-03-02
metadata scripts for UAI matching

mspector:


mspector:
a lot of this is consolidating and cleaning up old files (that's why I'm deleting so many things)

PR #271, opened 2017-03-09
Modify reviewer bids page to allow users to search subject area dropdown

zbialecki:


zbialecki:
I wasn't able to reproduce that js error, but I added the All option and did more testing and it seems to be working correctly. Could you try testing once more?

melisabok:
You are right, it was my data that didn't have the field TL:DR, I pushed a fix. Everything is working now.

PR #272, opened 2017-03-10
create workshop version of get-lazy-reviewer script

pmandler:


pmandler:
It's the same as get-lazy-reviewer.py except I replaced "conference" with "workshop" and removed a bit of unused code.

PR #279, opened 2017-03-15
Feature/rss init

pmandler:
Add nonreviewers group when submission uploaded.
Update processToFile to add TRACK_NAME to all functionTrack() calls.


iclr2017/python/assign-reviewers.py
```diff
@@ -1,145 +0,0 @@
-#!/usr/bin/python
```
melisabok: why are we removing this file?

rss2017/process/processToFile.js
```diff
@@ -0,0 +1,41 @@
...
+		var fileName = m.replace('<<', '').replace('>>', '');
+		var fileContent = fs.readFileSync(outputPath + '/' + fileName, 'utf8');
+		mainProcess = mainProcess.replace(m, fileContent);
+		console.log('Processing file', fileName);
+	})
+}
+
+
+// If there is track information add it as an argument to the function call
+if (track) {
```
melisabok: why are we doing this?

rss2017/python/superuser-init.py
```diff
@@ -0,0 +1,279 @@
...
+    proceeding_invitation = openreview.Invitation(PROCEEDINGS,
+        'Submission',
+        readers = ['everyone'],
+        writers = [PROCEEDINGS],
+        invitees = ['~'],
+        signatures = [PROCEEDINGS],
+        duedate = TIMESTAMP_DUE,
+        process = '../process/submissionProcessProceedings.js')
+
+    ## PAM why reply.copy instead of just reply?
```
melisabok: @pmandler you don't need to add you name in the comments.

rss2017/process/processToFile.js
melisabok: Please, avoid creation of a new script: rss2017/process/processToFile.js

try to solve it so we can use it in all the conferences. One possible solution could be:

```
function() {
   var or3client = lib.or3client;
   const TRACK_NAME = <<param1>>
   const CONFERENCE = 'roboticsfoundation.org/RSS/2017/Workshop';
   const COCHAIRS = CONFERENCE+"/Program_Co-Chairs"
   const TRACK = CONFERENCE+'/'+TRACK_NAME
   const PAPERGRP = TRACK+'/Paper' + note.number;
   const PAPERINV = TRACK+'/-/Paper' + note.number;
   const DUE_DATE = new Date(2017, 6, 30, 17, 15);

```

and then if the processToFile receives more than 3 arguments then find and replace:

```
mainProcess.replace('<<param1>>', process.argv[4])
```

in general

```
mainProcess.replace('<<param' + i + '>>', process.argv[3 + i])
```

rss2017/process/commentProcess.js
```diff
@@ -0,0 +1,34 @@
...
+      var forum = result.notes[0];
+      var note_number = forum.number;
+
+      var reviewers = [PAPERGRP + '/Reviewers'];
+      var authors = forum.content.authorids;
+
+      var author_mail = {
+        "groups": authors,
+        "subject": "Review of your submission to "+TRACK+": \"" + forum.content.title + "\"",
+        "message": "Your submission to " + conference + " has received a public comment.\n\n
```
melisabok: there is something weird in this string.

rss2017/process/commentProcess.js
```diff
@@ -0,0 +1,34 @@
...
+            Title: " + note.content.title + "\n\nReview: " + note.content.comment + "\n\n
+            To view the review, click here: " + baseUrl+"/forum?id=" + note.forum
+      };
+
+      var authorMailP = or3client.or3request( or3client.mailUrl, author_mail, 'POST', token );
+
+      return Promise.all([
```
melisabok: if there is only one promise you don't need to use Promise.all just return the single promise

rss2017/process/officialReviewProcess.js
```diff
@@ -0,0 +1,32 @@
...
+
+      var author_mail = {
+        "groups": authors,
+        "subject": "Review of your submission to "+TRACK+": \"" + forum.content.title + "\"",
+        "message": "Your submission to " + conference + " has received an official review.\n\nTitle: " + note.content.title + "\n\nReview: " + note.content.review + "\n\nTo view the review, click here: " + baseUrl+"/forum?id=" + note.forum
+      };
+
+      var authorMailP = or3client.or3request( or3client.mailUrl, author_mail, 'POST', token );
+
+      return Promise.all([
```
melisabok: same here:

```
return authorMailP
```

rss2017/process/submissionProcess.template
```diff
@@ -0,0 +1,139 @@
...
+      'forum': note.id,
+      'replyto': note.id,
+      'writers': {'values-regex': PAPERGRP+'/AnonReviewer[0-9]+'},
+      'signatures': {'values-regex': PAPERGRP+'/AnonReviewer[0-9]+'},
+      'readers': {
+        'values': ['everyone'],
+        'description': 'The users who will be allowed to read the above content.'
+      },
+      'content': {
+        'title': {
```
melisabok: Remember to add if the fields are required or not

rss2017/process/processToFile.js
pmandler: In RSS there are two tracks (like ICLR).  In order to avoid creating nearly identical copies of the process templates and js files I added the track parameter.  
I was nervous about adding it to the existing processToFile because I didn't want to mess up existing conferences. I like your solution.

rss2017/python/superuser-init.py
pmandler: I only add my name when it is a question.

iclr2017/python/assign-reviewers.py
pmandler: No idea.  I assume it was an accident.  How do I fix it?

mspector: I don't know the best way to do this with git - I'm not sure if Melisa knows, but I'm taking this opportunity to look it up

melisabok: you only need to commit the file again.

rss2017/process/processToFile.js
melisabok: Don't worry about messing things up, we always can fix them. And this is a good feature to add for other conferences. We should avoid duplicated code, it is difficult to maintain.

rss2017/process/commentProcess.js
melisabok: I guess there are some quotes missing here

scripts/processToFile.js
```diff
@@ -22,5 +24,29 @@ if(matches) {
...
 var newFileName = mainProcessFile.replace('.template', '.js');
+
+// Look for <<param>>, see if there is a index attached
+// and replace that string with the corresponding argument.
+//   <<param>> will be replaced w/ process.argv[PARAM_START_INDEX]
+//   <<param2>> will be replaced w/ process.argv[PARAM_START_INDEX+2]
+regex = new RegExp('<<(param.*)>>', 'g');
+matches = mainProcess.match(regex);
+if(matches) {
+    var param_max = process.argv.length - PARAM_START_INDEX;
```
melisabok: what about this, it is easier:

```
for(var i = 0; i < param_max; i++) {
  mainProcess = mainProcess.replace('<<param' + i + '>>', process.argv[PARAM_START_INDEX + i]);
}
```

it is not necessary to find the matchs of 'param.*'. The replace function doesn't do anything if it doesn't find the token.

melisabok: @pmandler could you do a refactor of this change and then we can merge this PR?
I think my suggestion is clearer and easier to implement.

pmandler: This only catches the first instance of param1 in the file.

melisabok: I think the python replace function replaces all the occurrences of param1, right?

pmandler: Not when I just tested it.  I can look around and see if there is a replaceAll or something.

pmandler: Probably python does, but js doesn't.

melisabok: Ohhh I'm sorry, it is javascript. You can use a global regex:

```
str.replace(new RegExp('<<param1>>', 'g'), 'value');
```

PR #280, opened 2017-03-16
Given the papers is not writable we need to add the authors as reader…

melisabok:
…s of the note


melisabok:
@mspector do you think it is ok to set the authorids and signatures?

mspector:
@melisabok yes, I think it's a good idea. I can't think of any negative consequences

PR #284, opened 2017-03-16
parametrize the name of the conference, by default is conference, for…

melisabok:
… workshop:

Fix divide by zero error

get-all-paper-status.py --name workshop


melisabok:
@pmandler I added a parameter so we can use the same script for the workshop:

```
get-all-paper-status.py --name workshop
```

PR #285, opened 2017-03-16
refactored precreate profiles so it can be used as a module

mspector:


uai2017/python/precreate_profiles.py
```diff
@@ -0,0 +1,144 @@
+import openreview
```
melisabok: can we move the script to the scripts folder ? it is not UAI conference related.

mspector: yes, good idea

PR #287, opened 2017-03-20
updated authorids description

mspector:


uai2017/python/superuser-init.py
```diff
@@ -222,7 +222,7 @@ def overwrite_allowed(groupid):
                 'required':True
             },
             'authorids': {
-                'description': 'Comma separated list of author email addresses, in the same order as above. Be sure each email address is linked to the each author profile.',
+                'description': 'Comma separated list of author email addresses, lowercased, in the same order as above. Be sure each email address is linked to the each author profile.',
```
melisabok: @mspector thanks Michael, could you fix this type please?

is linked to the each author profile

it should say: is linked to each author profile, right?

mspector: Ah, yes. Good catch.

melisabok:
Thanks!

PR #296, opened 2017-03-23
Fix/invitation constructor api

mspector:
formerly, the openreview.Invitation() constructor accepted two required arguments: one for the "owner" and one for the invitation name. Now it only accepts an Id, just like openreview.Group and openreview.Note


mspector:
corresponds with this pull request in openreview-py:
https://github.com/iesl/openreview-py/pull/39

PR #300, opened 2017-03-25
Update UAI Add/Bid webfield to save searches and filters

zbialecki:
- Add link to conference homepage in the status bar
- Add new script (similar to update-homepage.py) that just updates invitation webfields.


zbialecki:
I updated the webfield on dev, so it's testable there.

PR #311, opened 2017-04-01
disable invitation button

melisabok:


melisabok:
@mspector now you can merge it

PR #314, opened 2017-04-04
small bug fixes and notifications

mspector:


uai2017/python/metadata-papers.py
```diff
@@ -53,11 +54,13 @@ def prettify(elem):
 bids_by_number = defaultdict(list)
 bids_by_id = defaultdict(list)
 
+print "Processing bids... (this may take a while)"
 for b in bids:
     n = client.get_note(b.forum)
```
melisabok: why don't you get the notes from the submissions call? instead of doing call per bid

uai2017/python/metadata-papers.py
```diff
@@ -66,11 +69,13 @@ def prettify(elem):
 recs_by_number = defaultdict(list)
 recs_by_id = defaultdict(list)
 
+print "Processing recommendations..."
 for r in recs:
     n = client.get_note(r.forum)
```
melisabok: same here

PR #317, opened 2017-04-04
set the CONFERENCE group as reader of the tags

melisabok:
this depends on : https://github.com/iesl/openreview/pull/602


uai2017/python/superuser-init.py
```diff
@@ -534,14 +534,11 @@ def overwrite_allowed(groupid):
         'invitation': blind_submission_invitation.id,
         'readers': {
             'description': 'The users who will be allowed to read the above content.',
-            'value-regex': '~.*'
+            'values-copied': [CONFERENCE, '{signatures}']
```
mspector: should the areachairs group be here too?

PR #320, opened 2017-04-05
added .get() calls to get primary and additional areas

mspector:


uai2017/python/uai-match.py
```diff
@@ -288,8 +288,11 @@
     for reviewer in reviewers_group.members:
         if reviewer in registered_expertise_by_reviewer.keys():
             paper_subjects = paper_note.content['subject areas']
-            primary_subjects = registered_expertise_by_reviewer[reviewer]['primary area']
-            secondary_subjects = registered_expertise_by_reviewer[reviewer]['additional areas']
+            primary_subjects = registered_expertise_by_reviewer[reviewer].get('primary area')
```
melisabok: with the get function you can set a default value as second parameter:

```
registered_expertise_by_reviewer[reviewer].get('primary area', [])
```

mspector: ah, great! thanks!

PR #327, opened 2017-04-10
Feature/efficiency improvements

mspector:


uai2017/python/uai-metadata.py
```diff
@@ -0,0 +1,174 @@
...
+
+import argparse
+import csv
+import sys
+import json
+import numpy as np
+from collections import defaultdict
+
+import openreview
+import openreview_matcher
```
melisabok: could you remove this dependency?

mspector: just removed all the extraneous dependencies

melisabok:
if I run:

```
python uai-metadata.py --username OpenReview.net --password OpenReview_beta --upload metadata.pkl
```

get:

```
IOError: [Errno 2] No such file or directory: 'metadata.pkl.pkl'
```

in the help says:

```
  --upload UPLOAD      the .pkl file (with extension) to upload to OpenReview.
                       If "true", defaults to ./metadata.pkl
```

uai2017/python/match_utils.py
```diff
@@ -1,3 +1,13 @@
+import pickle
+
+def save_obj(obj, name):
+    with open(name + '.pkl', 'wb') as f:
+        pickle.dump(obj, f, pickle.HIGHEST_PROTOCOL)
```
mspector: I'm leaving the .pkl extension in the save_obj function here because the user shouldn't be able to save a non-pkl file.

unfortunately this means that save/load are asymmetrical, which is strange.

is this an OK way to do it? on the one hand, it's nice to pass in autocompleted full filenames, but on the other hand, we don't want people trying to save non-pkl files

melisabok: well you can accept only pkl files.

PR #331, opened 2017-04-11
don't overwrite groups if they already exist

mspector:
sorry for the huge diff, it's mostly spacing issues. there was something weird with the spacing/tabs

### DIFF
--- a/uai2017/python/create-invitations.py
+++ b/uai2017/python/create-invitations.py
@@ -20,268 +20,271 @@ maskAuthorsGroup = maskPaperGroup + "/Authors"
 
 def get_open_comment_invitation(submissionId, number, authorsGroupId):
 
-	allGroups = [COCHAIRS, PC, SPC, authorsGroupId]
-	reply = {
-    	'forum': submissionId,
-      	'signatures': {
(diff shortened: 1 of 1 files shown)


uai2017/python/create-invitations.py
```diff
@@ -301,57 +304,68 @@ def get_meta_review_invitation(submissionId, number, authorsGroupId, areachairGr
...
+
+    paper_group = client.get_group(id = maskPaperGroup.replace('[PAPER_NUMBER]', str(submission.number)))
+
+    if paper_group:
+
+      author_group = client.get_group(id = maskAuthorsGroup.replace('[PAPER_NUMBER]', str(submission.number)))
+
+      if author_group:
+        # Check the necessary groups and post them if they don't exist
+        if not client.exists(paper_group.id + '/Reviewers'):
```
melisabok: exists function is doing a get call and then you do an extra get_group. 

I suggest to do a get_group directly and if we get an error do the post:

```
reviewers_group = client.get_group(paper_group.id + '/Reviewers')
if not reviewers_group:
      .. do the post call
```

melisabok:
thanks Michael! I'm sorry for the indentation issue.

PR #333, opened 2017-04-12
add conflict detection to assignment scripts

mspector:
for each of "assign-program-committee" and "assign-senior-program-committee", the script now finds all the conflict domains for the authors of the paper, and adds those domains as nonreaders to the groups Paper#/Reviewers, Paper#/Area_Chair and Paper#/AnonReviewer#

it also alerts you if you try to assign a reviewer or area chair to a paper with which they have a conflict, and gives you the option to either remove the conflict and continue, or to cancel the assignment (and move on, if you're assigning a list of users). This caused many problems for ICLR, because the user was not a reader of their own Reviewer or Area Chair group.

### DIFF
--- a/uai2017/python/assign-program-committee.py
+++ b/uai2017/python/assign-program-committee.py
@@ -17,7 +17,7 @@ from uaidata import *
 # Argument handling and initialization
 # .............................................................................
 parser = argparse.ArgumentParser()
-parser.add_argument('-a','--assignments', help="either (1) a csv file containing reviewer assignments or (2) a string of the format '<openreview_id>,<paper#>' e.g. '~Alan_Turing1,23'")
+parser.add_argument('-a','--assignments', help="either (1) a csv file containing reviewer assignments or (2) a string of the format '<openreview_id>,<paper#>' e.g. '~Alan_Turing1,23'", required=True)
 parser.add_argument('-o','--overwrite', help="if true, erases existing assignments before assigning")
 parser.add_argument('--baseurl', help="base url")
 parser.add_argument('--username')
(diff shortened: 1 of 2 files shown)


uai2017/python/assign-program-committee.py
```diff
@@ -49,7 +49,25 @@ def single_assignment_valid(s):
     except IndexError:
         return False
 
-def assign_reviewer(reviewer,paper_number):
+def get_nonreaders(paper_number):
+    # nonreaders are:
+    # (1) all authors of the paper
+    # (2) all domain groups of the authors of the paper
```
melisabok: I think adding only the domains is enough, right?

uai2017/python/assign-program-committee.py
```diff
@@ -137,22 +159,42 @@ def clear_assignments():
...
-        print "Invalid input"
-        sys.exit()
+
+def assign(assignments):
+    for a in assignments:
+        reviewer = a[0]
+        paper_number = a[1]
+        conflict_list = get_nonreaders(paper_number)
+
+        members = set([g.id for g in client.get_groups(member=reviewer)])
```
melisabok: I like this!

uai2017/python/assign-senior-program-committee.py
```diff
@@ -49,7 +49,25 @@ def single_assignment_valid(s):
     except IndexError:
         return False
 
-def assign_areachair(areachair, paper_number):
+def get_nonreaders(paper_number):
```
melisabok: we need to start to think about the duplicated code.

uai2017/python/assign-program-committee.py
mspector: oops, the comment is out of date. it does only do domains

uai2017/python/assign-senior-program-committee.py
mspector: yes, agreed :/

PR #336, opened 2017-04-13
Fix/assignment scripts

mspector:
should there be a --overwrite flag? or is --add and --remove enough?

### DIFF
--- a/uai2017/python/assign-program-committee.py
+++ b/uai2017/python/assign-program-committee.py
@@ -110,7 +110,7 @@ def create_reviewer_group(new_reviewer_id, reviewer, paper_number, conflict_list
         signatures=['auai.org/UAI/2017'],
         writers=['auai.org/UAI/2017'],
         members=[reviewer],
-        readers=[CONFERENCE, COCHAIRS, SPC, PC],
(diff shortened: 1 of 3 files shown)


uai2017/python/assign-program-committee.py
```diff
@@ -135,34 +104,22 @@ def get_reviewer_group(reviewer, paper_number, conflict_list):
...
 
-
 def create_reviewer_group(new_reviewer_id, reviewer, paper_number, conflict_list):
     new_reviewer = openreview.Group(
         new_reviewer_id,
         signatures=['auai.org/UAI/2017'],
         writers=['auai.org/UAI/2017'],
         members=[reviewer],
-        readers=[CONFERENCE, COCHAIRS, SPC, PC],
+        readers=[CONFERENCE, COCHAIRS, 'auai.org/UAI/2017/Paper%s/Area_Chair', 'auai.org/UAI/2017/Paper%s/Reviewers'],
```
melisabok: how do you set the paper number here?

mspector: ah!! good catch.

PR #338, opened 2017-04-14
Fix/uai invitation types

melisabok:


uai2017/python/create-invitations.py
```diff
@@ -23,8 +23,9 @@ def get_open_comment_invitation(submissionId, number, authorsGroupId):
     allGroups = [COCHAIRS, PC, SPC, authorsGroupId]
     reply = {
         'forum': submissionId,
+        'selfReplyOnly': True,
         'signatures': {
-            'values-regex': '|'.join(allGroups),
+            'values-regex': '|'.join(allGroups) + '|' + CONFERENCE + '/Paper' + str(number) + '/AnonReviewer[0-9]+',
```
mspector: I would move CONFERENCE + '/Paper' + str(number) + '/AnonReviewer[0-9]+' to the allGroups list, just to keep things consistent... what do you think?

melisabok: I'm using allGroups in other places and they are a list of real groups not regex. 

Look at the readers values

mspector: oh I see. ok

PR #345, opened 2017-04-17
Fix / Lodash version upgrade

zbialecki:
Update all webfields to not load older version of lodash and make sure all uses of _ are compatible with version 4.17

Updated venues:
ICLR 2016
NIPS 2016 DLSymposium
NIPS 2016 MLITS
UAI 2017
RSS 2017

### DIFF
--- a/iclr2016(old scripts)/20160114-demo/web-field-workshop.html
+++ b/iclr2016(old scripts)/20160114-demo/web-field-workshop.html
@@ -1,8 +1,4 @@
 <html>
-  <head>
-    <!--<script src="https://code.jquery.com/jquery-1.10.1.min.js"></script>-->
-    <script src="https://cdn.jsdelivr.net/lodash/3.5.0/lodash.compat.min.js"></script>
-  </head>
   <body>
     <div id='main'>
       <div id='header'></div>
--- a/iclr2016/iclr2016_webfield.html
+++ b/iclr2016/iclr2016_webfield.html
@@ -1,8 +1,4 @@
 <html>
-  <head>
-    <!--<script src="https://code.jquery.com/jquery-1.10.1.min.js"></script>-->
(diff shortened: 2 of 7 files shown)


iclr2016(old scripts)/20160114-demo/web-field-workshop.html
```diff
@@ -1,8 +1,4 @@
 <html>
-  <head>
-    <!--<script src="https://code.jquery.com/jquery-1.10.1.min.js"></script>-->
-    <script src="https://cdn.jsdelivr.net/lodash/3.5.0/lodash.compat.min.js"></script>
```
mspector: another naive question: I understand that removing this code stops the file from loading the older version of lodash, but where does the newer version come from, now that this is gone?

zbialecki: As it works now, once the webfield is loaded via ajax, cajaManager appends the html directly to the dom in either the div#group-container or div#invitation-container depending on the page. Once this happens all scrips are evaluated directly in the context of the current page, with access to all global variables, including $, _, Handlebars.templates, and Open Review globals like controller and view.

This obviously isn't super safe...

zbialecki: This is part of the reason I'd like to move away from the webfields being full HTML files. Since they aren't being loaded in an iframe (that would be a whole other set of headaches), we really only need them to be .js files, which are easier to sanitize, validate, and execute in a sandbox environment where we control what global variables are accessible.

melisabok: Sounds good to me. Those are old webfields, the new ones don't include any js library.

zbialecki:
I think this is ready to be merged. Is there anything else holding it back?

PR #349, opened 2017-04-19
remove unnecessary call to get tags

melisabok:


melisabok:
Yes, bids and recommend reviewers are closed now. @mspector could you merge it?

PR #350, opened 2017-04-19
Fix review invitation name

melisabok:


melisabok:
We changed the name of the review invitation by request of the pcs

PR #357, opened 2017-04-24
reorganized

mspector:
most of the changes are renames.

I deleted some ICML files that didn't seem necessary. Some were related with the legacy conference, which should be unnecessary now that we've migrated over.

I also split the ICML workshops into two directories, to reflect their differing paths

this reorg moved the processToFile.js script, so I edited the superuser-init functions for our currently active conferences so that they can find it.

finally, I deleted some data files that I don't think should be stored in the public repo


venues/auai.org/UAI/2017/python/superuser-init.py
```diff
@@ -10,11 +10,15 @@
 ## Import statements
 import argparse
 import csv
-import sys
 import openreview
 from uaidata import *
 from subprocess import call
 
+import sys, os
+def get_path(path): return os.path.join(os.path.dirname(__file__), path)
```
mspector: I'm not refactoring get_path into utils because finding the path of the executing script (not the module file utils.py) might not be possible. See the answer here:

http://stackoverflow.com/questions/2632199/how-do-i-get-the-path-of-the-current-executed-file-in-python

PR #359, opened 2017-04-27
Fix/script generalization

mspector:
This is a first pass at script generalization

run generate-conference.py to see how it works


melisabok:
Here is my first feedback:

- When I run generate-conference it doesn't ask me for user/password, is this script runnable only for the super user?
- It doesn't ask me for a password for the admin user, I'm not sure where to create this user: here or in the superuser-init. But it would be nice that the superuser-init can be run by admin user, what do you think? 
- When I run the superuser-init with the super user I get:

```
{ name: 'error',
  errors:
   [ { type: 'forbidden',
       path: 'id',
       value: 'melisabok.org/ICML/2017',
       user: 'OpenReview.net' } ] }
```

And I think it is because you need to create all the groups: melisabok.org, melisabok.org/ICML and melisabok.org/ICML/2017.

mspector:
@melisabok thanks for the comments!

generate-conference doesn't interact with the server at all, so it can be run by anyone

regarding the admin user: I was thinking that it might be useful to generate two scripts: "superuser-init", and "init". superuser-init would create the admin and all the parent groups that don't already exist. init would create all the groups for that conference, and it would be runnable by the admin.

mspector:
@melisabok @mzarozinski 

no rush, but this is ready for another test, if either of you want to give it a try.

PR #360, opened 2017-05-02
Display min/max rating and review length

melisabok:


melisabok:
thanks @mspector , please don't merge it. I will use the same PR for all the changes

PR #362, opened 2017-05-04
Feature/uai console sorting

melisabok:
Adding a sort panel in each tab to sort the data in the client side. 

Each tab has different sort options. 

@zbialecki I reused some classes of the search panel that we have in the bid webfield. I think it is ok to leave them otherwise the panel is broken :(. I also had to overwrite some styles. 

It is deployed in dev and live, please take a look.


zbialecki:
I just pushed some style fixes, I think it should look better now.

screen shot 2017-05-04 at 8 39 01 pm

melisabok:
Thanks @zbialecki , I have an issue with the sort order since it is a button. When I click the button to change the order it tries to refresh the page and goes to the home page. 

It is deployed on dev.

zbialecki:
Did the last deploy also include my second commit? It's probably something silly like I didn't return false at the end of the click handler.

PR #364, opened 2017-05-08
fix meta review form fields

melisabok:
@mspector : what do you think about the not required fields? if the recommendation is 'Reject' the next fields shouldn't be mandatory.

### DIFF
--- a/venues/auai.org/UAI/2017/python/create-invitations.py
+++ b/venues/auai.org/UAI/2017/python/create-invitations.py
@@ -297,18 +297,11 @@ def get_meta_review_invitation(submissionId, number, authorsGroupId, areachairGr
             'recommendation': {
                     'order': 3,
                     'value-radio': [
-                        'Accept',
+                        'Accept (Oral)',
+                        'Accept (Poster)',
                         'Reject'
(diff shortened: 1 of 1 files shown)


mspector:
I'm not sure about the format field... If they select "accept" in the recommendation field, then it might be important that they specify which kind of acceptance it is.

Is there a default value for the "values-radio" widget?

I agree that the "best student paper" field can be optional.

melisabok:
Yes, I think the same. but it is weird that the recommendation is 'reject' and ask for a format.

mspector:
If values-radio allows default values, that could solve the problem.

Another solution is to make the format field required, but provide a "Reject" option there too

Finally, we could just assume that any reviews with an unfilled format field is a "poster" format. We would have to let Gal and Kristian know of this, but I think it would be ok.

melisabok:
Or we can merge recommendation and format in a single field:

```
            'Accept (Oral)',
            'Accept (Poster)',
            'Reject',
```

same as ICLR. Keep best paper fields optional.

mspector:
Ok, that's a good solution

PR #368, opened 2017-05-11
save the last reminder sent in the local storage and display it below…

melisabok:
… 'Send Reminder'

this is a temporal solution to display the last time the reminder was sent so the area chair doesn't send the reminder twice. 

I had to refactor the render of the table so we can refresh the view after the reminder was sent. But I had to add a clone deep of one of the fetched data because mysteriously the values were undefined.

here is an attachment of how it looks:
screen shot 2017-05-11 at 3 41 31 pm


melisabok:
Related to PR: https://github.com/iesl/openreview/pull/695

mspector:
the interface looks good to me!

PR #369, opened 2017-05-12
Feature / Webfield enhancements

zbialecki:
Add new webfield templates based on completely new Webfield library in https://github.com/iesl/openreview/pull/697

These templates can be used as a base to customize for other conferences. Initially I created 2 - one based on the ICLR.cc/2017/conference group and one based on the auia.org/UAI/2017 group.

Due to some css changes, the UAI webfields should be updated after the other PR is merged.


melisabok:
The new UI looks very nice! I really like it: I love the new header, I like the pagination and the search. This is really helpful. I hope this could be more understandable for the program chairs. 

@mspector I made some changes in the templates so we can use the new webfield. Please, take a look. 

One little comment: we should display the search textbox when there is not submitted papers. If you try to search it throws an error and goes to the homepage. 
screen shot 2017-05-12 at 4 30 21 pm

I also tested the webfield with tabs, this is a good option to have and it can be integrated in the generate conference scripts, so depending of the kind of conference we can use one webfield or the other. I think the webfield with tabs could be used for 'double-blind' conferences, and the basic webfield for conferences where all the papers and reviews are open. @mspector what do you think?

zbialecki:
@melisabok Glad you like it! Good catch with the search not working when there aren't any notes, I'll fix it in the other PR.

I think it would be cool to use the tabbed page as a base for blind submission conferences, but I will say it requires more customization than the other template. Right now there are still some UAI-specific regexes in the code, the subject area list would need to be updated per conference, and there might be 1 or 2 other things.

melisabok:
Yes, you are right. generate-conference.py can ask for that information and subject areas also. I think we can add it in the version 2 of the script.

melisabok:
@mspector do you agree with the changes, we can merge it?

mspector:
this looks really great. thanks @zbialecki !

zbialecki:
Awesome, glad you like the new API Michael. One small thing, when this gets deployed on openreview it would be good to update the UAI webfields as there were some CSS changes. It's nothing breaking, but it will look better.

melisabok:
I just made a deploy of openreview in dev. Should I update only UAI webfields?

zbialecki:
Yeah, I think for testing we should update those. The only ones that are affected are the home page, the PC console page and the SPC console page.

melisabok:
Ok, I updated the three webfields, could you check that are ok?

zbialecki:
They look good, except that when I try to view the SPC console http://dev.openreview.net/group?id=auai.org/UAI/2017/Senior_Program_Committee the POST /user/profiles api call returns the error 'Invalid value for ids:'

It may be that I'm not logged in as the right user but I'm not sure.

melisabok:
Yes, you need to login as uaiareachair@openreview.net

zbialecki:
OK, that worked. The SPC console looks good too.

melisabok:
Ok, it is deployed in live.

PR #372, opened 2017-05-15
Add TIMESTAMP to config file.  Change how the due date is checked for…

pmandler:
… validity. Use the more thorough testing done in datetime function (called by utils.get_duedate()).

### DIFF
--- a/admin/generate-conference.py
+++ b/admin/generate-conference.py
@@ -6,6 +6,7 @@ import utils
 import templates
 import argparse
 import datetime
+import time
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--test', help="base URL")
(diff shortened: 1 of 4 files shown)


pmandler:
Zach added timestamp to the webfield.  I added the timestamp to the config.py file. I also changed the way it tests for a valid time/date which is what appeared to be a conflict.

admin/generate-conference.py
```diff
@@ -19,34 +20,30 @@
...
+# check if due date/time is valid and after current time
+duedate_input = raw_input("Enter the duedate (DD/MM/YYYY): ")
+try:
+	day, month, year = duedate_input.split('/')
+	day = int(day)
+	month = int(month)
+	year = int(year)
+	# this will catch invalid values for date
+	duedate = utils.get_duedate(year, month, day)
+	current_timestamp = int(round(time.time() * 1000))
```
melisabok: you can get the current date using this:
```
now = datetime.datetime.now()
```

and then compare python dates:

```
duedate > now
```

PR #375, opened 2017-05-15
Feature/matching redesign

mspector:
should be merged simultaneously with:

https://github.com/iesl/openreview-matching/pull/3


venues/auai.org/UAI/2017/js/reset-metadata.js
```diff
@@ -0,0 +1,31 @@
+// Run this script when the UAI paper metadata needs to be reset
+
+var notes = db._query("FOR note IN openreview_notes FILTER CONTAINS(note.invitation, 'auai.org/UAI/2017/-/Paper/Metadata') RETURN note._id").toArray()
```
melisabok: Are you going to use the new endpoint?

If not, you can do a removeByExample({ invitation: 'auai.org/UAI/2017/-/Paper/Metadata'}) and don't need to iterate through all the notes.

PR #377, opened 2017-05-18
get all the reviews using pagination

melisabok:


venues/auai.org/UAI/2017/python/email-lazy-reviewers.py
```diff
@@ -55,12 +55,31 @@
 total_missing = 0;
 total_complete = 0;
 
+
+def get_reviews():
+
+    invitation_id = 'auai.org/UAI/2017/-/Paper.*/' + invitation
```
mspector: this variable is unused

PR #384, opened 2017-05-24
add default process functions for submission and comments

mspector:
the only non-obvious change here is that I have changed the generate-conference script to write the webfield in the same way that the other files are written. I don't think we should use the "template" module to build the webfields anymore.

I will create a new pull request to refactor out the template module, but this will work for Pam trying to generate the ICML workshop tomorrow

### DIFF
--- a/admin/conference-template/process/submissionProcess.template
+++ b/admin/conference-template/process/submissionProcess.template
@@ -1,0 +1,19 @@
+function () {
+  var or3client = lib.or3client;
+
+  var CONFERENCE = <<CONF>>;
+  var PAPERINV = CONFERENCE+'/-/Paper' + note.number;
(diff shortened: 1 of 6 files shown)


admin/conference-template/process/submissionProcess.template
```diff
@@ -0,0 +1,19 @@
+function () {
+  var or3client = lib.or3client;
+
+  var CONFERENCE = <<CONF>>;
+  var PAPERINV = CONFERENCE+'/-/Paper' + note.number;
```
melisabok: CONFERENCE and PAPERINV are not being used, we can remove them

melisabok:
Thanks @mspector , very nice job!

PR #397, opened 2017-05-30
Fix/dblp import

mzarozinski:
Updated scripts to get around "special case" titles. This version writes out an error log for any records that were not successfully saved.

### DIFF
--- a/venues/DBLP.org/python/dblp_wrapper.py
+++ b/venues/DBLP.org/python/dblp_wrapper.py
@@ -1,0 +1,66 @@
+#!/usr/bin/python
+
+"""
+
+This is the initialization script for dblp.org
+added dblp directory and tutorial section
+
+It should only be run ONCE to kick off the conference. It can only be run by the Super User.
+
+"""
+
+## Import statements
+import argparse
+import csv
+import sys
+from openreview import *
+import update_records
+import json
+import os
+import traceback
+
+## Handle the arguments
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--overwrite',
+                    help="If set to true, overwrites existing groups")
+parser.add_argument('--username')
+parser.add_argument('--password')
+parser.add_argument('--json')
+
+args = parser.parse_args()
+## Initialize the client library with username and password
+if args.username != None and args.password != None:
+    openreview = Client(baseurl=args.baseurl, username=args.username, password=args.password)
+else:
+    openreview = Client(baseurl=args.baseurl)
+
+dblp_inv = openreview.get_invitation('DBLP.org/-/paper')
+
+data = json.loads(open(args.json).read())
+
+count = 0
+for d in data:
+    count += 1
+    try:
+        ## Not all data from DBLP exactly fits the invitation we've defined, so we need to rename a few fields.
+        content = update_records.process_content(d)
+
+        ## Use the post_or_update function to post a new record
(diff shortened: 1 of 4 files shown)


venues/DBLP.org/python/dblp_wrapper.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/python
+
+"""
+
+This is the initialization script for dblp.org
+added dblp directory and tutorial section
+
+It should only be run ONCE to kick off the conference. It can only be run by the Super User.
+
+"""
```
mspector: could you change this docstring (or remove it entirely) and rename the file to something more descriptive?

maybe "dblp_upload.py", or something like that?

venues/DBLP.org/python/dblp_wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+import traceback
+
+## Handle the arguments
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--overwrite',
+                    help="If set to true, overwrites existing groups")
+parser.add_argument('--username')
+parser.add_argument('--password')
+parser.add_argument('--json')
```
mspector: a "help" string for the json argument might be useful, too

venues/DBLP.org/python/dblp_wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+        f.write("line: " + str(count) + " : " + str(d) + os.linesep)
+        traceback.print_tb(exc_traceback, None, file=f)
+        f.write(os.linesep)
+        f.close()
+
+    print count
+
+
+
+pass
```
mspector: what is this "pass" for?

venues/DBLP.org/python/superuser-init.py
```diff
@@ -112,7 +112,7 @@ def overwrite_allowed(groupid):
                 'description': 'Comma separated list of author email addresses, in the same order as above.',
                 'order': 3,
                 'values-regex': "[^;,\\n]+(,[^,\\n]+)*",
-                'required':True
+                'required':False
```
melisabok: Why this field is not required?

mzarozinski: When we add a revision to an existing record, if this is required the ["UNK"] values are inserted which show up for the paper rather than the existing (good) data already in the database.

melisabok: this should applies to the 'DBLP.org/-/Add/Revision' invitation reply only, not for the 'DBLP.org/-/paper'. 

BTW DBLP.org/-/paper should be renamed to DBLP.org/-/Submission, @mspector  what do you think?

mzarozinski:
I also need to remove the process function from superuser-init

venues/DBLP.org/python/superuser-init.py
mspector: I'm not sure. It's technically not a "submission". It might be a good idea to name it "DBLP.org/-/Upload".

It's not clear to me, since we haven't fully explored this use case for the system, but that would be what I would call it...

mzarozinski: do we have a decision on the invitation name? I'd side with "Upload".

melisabok: Ok, let's go with Upload then

PR #400, opened 2017-05-31
University of Rostock - initial information for course

pmandler:


melisabok:
Looks good to me!

venues/informatik.uni-rostock.de/Informatik_Rostock/Ubiquitous_Computing/2017/python/config.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/python
+
+import sys, os
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "/Users/mandler/projects/openreview-scripts/utils"))
```
mspector: Hey Pam - can you change this path (and others like it) to relative paths? 

Sorry about this. This is actually my fault - the generator script doesn't handle this properly, and needs to be refactored. It's on my plate to fix this, but for now, we just need to manually change these "utils" paths to be relative, so that others can run superuser-init if needed

pmandler:
Good catch.  Again.  Thanks.  
I got it for RML, but missed in in WHI so I’ll fix it there too.

Pam

PR #404, opened 2017-06-01
ICML MLAV Workshop - script to add groups and review invite when subm…

pmandler:
…issions done
submissions_done.py - new script to add groups and review invitation for each submission
config.py - add review due date
utils/templates.py - add review invitation template ('short' because offering 5 options instead of 10)
officialReviewProcess.js - debug


venues/ICML.cc/2017/MLAV/python/submission_done.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/python
```
melisabok: shouldn't be called create_invitations.py?

pmandler: Or create_groups?  Is there a naming convention for scripts?

melisabok: Well, in UAI I called create_invitations.py. That's was the final goals besides you need to create a bunch of groups to setup the invitation.  But we can find another more descriptive name.

melisabok: Here is another case: https://github.com/iesl/openreview-scripts/blob/master/venues/swsa.semanticweb.org/ISWC/2017/DeSemWeb/python/create-invitations.py

utils/templates.py
```diff
@@ -261,3 +261,55 @@ def __init__(self, params=None, content=None):
 
         self.body = copy.deepcopy(self.params)
         self.body.update({'content': self.content})
+
+class ShortReviewReply(InvitationReply):
+    """
```
mspector: This looks good, but I would like to (very soon) remove the whole /utils directory (the openreview-scripts repo isn't really designed to handle modules like this, and it's causing problems). Could you move this to the conference's config.py file?

pmandler: Ok for now.  
On the other hand, if we want to make this accessible to others, we shouldn't have to rewrite invitations from scratch.  Maybe this submissions_done (or create_invitation) script belongs in admin/conference-template.

venues/ICML.cc/2017/MLAV/python/submission_done.py
pmandler: I'll change it to create_invitiations.py to make it easier to find.
A line or two of comments at the top of scripts would make it a lot easier for me to figure out what they are used for. Should I be nervous when I see 'Not to spec' written at the top?

utils/templates.py
mspector: I think the longer term solution is actually to change the invitation options in the backend, to allow us to create a single invitation for reviews that is attached to all invitations of a certain ID. In the most basic conference, this will remove the need for the "create_invitations" script entirely.

it still might be a good idea to add a "create_invitations" template to the script generator, but it may not be necessary for the most basic conference configuration

melisabok: I agree, but the idea is that the PC can extend the basic configuration and create their own invitations.

melisabok: I guess the commit of this file should be reverted right?

mspector: I think so

mspector: actually, for simplicity, let's just leave it here. it will be removed when i make my other pull request.

@melisabok @pmandler is that ok?

melisabok: But I think @pmandler  is not using ShortReviewReply anymore. She moved the code to the config.py

pmandler: I think I should remove my change.  If I do:
git rm template.py
that will just remove it from the branch and not from the master - correct?

PR #411, opened 2017-06-09
assign-reviewer.py script with comments

pmandler:
Looks like there are additions in ISWC that I may want to grab.

### DIFF
--- a/samples/python/assign-reviewers.py
+++ b/samples/python/assign-reviewers.py
@@ -1,0 +1,202 @@
+#!/usr/bin/python
+
+###############################################################################
+''' Assigns reviewers to papers
+ ex. python assign-reviewers.py --baseurl http://localhost:3000
+       --username admin --password admin_pw 'reviewer@gmail.com,3'
+
+ Checks paper number valid.
+ Check reviewer email address or domain is not on the conflicts list.
+ Check reviewer is in the system.
+ If reviewer is not in conference reviewers group (config.CONF/Reviewers), add it.
+ If reviewer not already assigned to this paper:
+	Determine AnonReviewer number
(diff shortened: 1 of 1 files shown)


samples/python/assign-reviewers.py
```diff
@@ -0,0 +1,202 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument('assignments', help="either (1) a csv file containing reviewer assignments or (2) a string of the format '<email_address>,<paper#>' e.g. 'reviewer@cs.umass.edu,23'")
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+args = parser.parse_args()
+
+## Initialize the client library with username and password
+if args.username!=None and args.password!=None:
+    client = Client(baseurl=args.baseurl, username=args.username, password=args.password)
```
mspector: you should be able to remove the "if/else" block here, and just have a single line like this one

mspector:
Which additions in IWSC are you referring to? Some of the code in the scripts are specific to DeSemWeb, and won't necessarily apply to other conferences...

samples/python/assign-reviewers.py
pmandler: I just saw the extra arguments that allowed for to add or remove.  And it looks like it handles user id's or emails.  This one is specific to email addresses.

mspector: Ok, I see. Do you want to add those parameters before we approve this PR?

pmandler: I think later.  Unlike that script which puts all the reviewers into one AnonReviewer group, this one would have to find the AnonReviewer# to remove it from the Reviewers group.

PR #413, opened 2017-06-14
sample python new scripts

pmandler:
Basic scripts:
freeze-submissions.py: Make submissions non-editable / non-deletable 
get-submissions.py: Output submission info for assigning reveiwers 
create-invitations.py: create subgroups and reviewer invitations

They all need to be run in the same directory as config.py for the conference.  They do not use utils.py.

### DIFF
--- a/samples/python/create-invitations.py
+++ b/samples/python/create-invitations.py
@@ -1,0 +1,106 @@
+#!/usr/bin/python
+
+###############################################################################
+# ex. python create-invitations.py --baseurl http://localhost:3000
+#       --username admin --password admin_pw
+#
+# To be run after submission due date to create review invitations for all the papers.
+# For each paper:
+# 1) create authorGroup (can see reviews, can't write a review)
+#           reviewer group (reviewers for this paper)
+#           and nonReviewerGroup (folks that aren't allowed to read the review at least not yet)
+# 2) create review invitation
+###############################################################################
+
+## Import statements
+import argparse
+import config
+from openreview import *
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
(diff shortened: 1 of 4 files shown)


pmandler:
Tested in my local copy of MLAV. There shouldn't be anything specific to that workshop.

samples/python/create-invitations.py
```diff
@@ -0,0 +1,106 @@
...
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+args = parser.parse_args()
+
+## Initialize the client library with username and password
+if args.username!=None and args.password!=None:
+    client = Client(baseurl=args.baseurl, username=args.username, password=args.password)
+else:
+    client = Client(baseurl=args.baseurl)
```
mspector: in all these scripts, you can replace this if/else block with just a single line:

client = Client(baseurl=args.baseurl, username=args.username, password=args.password)

the Client object will search the environment variables for the username and password if they are None

pmandler:
Yes. And have the option to tweak things for each conference.  
If you think it’s better I could add an argument to get the conference directory.

pam

mspector:
I think a conference argument is a good idea (since it doesn't prevent people from copy/pasting it if they want)

melisabok:
We should be careful the program chairs can change data of other conferences, have we tested that?

pmandler:
I’m testing out the freeze-submissions script with different users and adding catches for the errors.  I have found that admin can edit the submission invitation, but does not have permission to edit the submission notes.  I need the change the submission note writers to ‘freeze’ the paper.  This makes me wonder if maybe this is a superuser function after all, not something we want to share with admins.  Getting paper info and creating review invitations I have no problem with.  But do we want admins to be able to edit a submission note?

Pam

PR #415, opened 2017-06-14
desemweb tweaks for initial conference opening

mspector:


mspector:
@pmandler - since @melisabok  is leaving soon, it seems that it is time for you to start reviewing my python scripts :)

pmandler:
When you moved the creation of the anonymous readers to the create-groups.py, why did you change the readers to be only ADMIN, not PROG_CHAIRS (line 38)?

When formatting strings is there a reason to use .format instead of %s?

mspector:
1. (I'm assuming you mean "members", not "readers"). I removed the program chairs because there is a possibility that the program chairs could be authors. When the program chairs are considered members of the anonymous reviewer group, even including them in the "nonreaders" group doesn't seem to stop them from seeing the poster's real identity. I don't know if this is a bug in the backend or not, but this was a quick way to fix it, and makes sense to me (i.e. why should program chairs automatically be reviewers? they should be assigned individually like everyone else)

2. I'm not sure, but I read somewhere that the %s way of doing things is going to be deprecated at some point. they recommended that people use "{0} ... {1} ...".format(firstArg, secondArg) instead

PR #416, opened 2017-06-15
Fix/improve admin workflow

mspector:


admin/superuser-init.py
```diff
@@ -60,95 +60,96 @@ def get_input_or_data(name):
...
-print "writing %s/process/commentProcess.js" % directory
-with open(directory + '/process/commentProcess.js', 'w') as new_commentprocess, open(utils.get_path('./conference-template/process/commentProcess.template', __file__)) as template_commentprocess:
-    templatestring = template_commentprocess.read().replace('<<PHRASE>>', "\"%s\"" % conference_phrase)
-    new_commentprocess.write(templatestring)
-
-print "writing %s/params.json" % directory
-with open(directory + '/params.json', 'w') as paramsfile:
-    json.dump(data, paramsfile, indent=4)
+
+if args.overwrite or not os.path.exists(directory):
```
melisabok: where do you create the root group now? I don't see the call

mspector: it gets created on line 176

melisabok: I'm running the superuser-init for the semantic web conference and I get an error trying to create the admin user:
screen shot 2017-06-16 at 10 09 44 am


I'm using the suggested admin id

mspector: sorry, the "suggestion" isn't a "default" - you would still have to enter the username you want. I can change this.

melisabok: Oops, I'm sorry.

Now it is working fine. I could create the semantic web conference.

PR #423, opened 2017-06-23
Improve basic webfield template for case when invitation is private

zbialecki:
This is a fairly minor change - most of the code changes are in the webfield module. It might be good to update the Informatik Rostock Ubiquitous Computing 2017 webfield with this change though.


melisabok:
I tested the Rostock conference without this change and it is working fine. Am I doing something wrong?

zbialecki:
No, I think it's OK. The change is really just an optimization so that if the invitation doesn't exist it doesn't try to render the button.

melisabok:
Sounds good to me !

PR #427, opened 2017-06-28
WHI allow reviews to be any length and spell out size limit on title.

pmandler:
If this looks good, let me know and I'll add the same change for the other conferences in the same pull.

### DIFF
--- a/venues/ICML.cc/2017/WHI/python/config.py
+++ b/venues/ICML.cc/2017/WHI/python/config.py
@@ -100,12 +100,12 @@ review_content = {
     'title': {
         'order': 1,
         'value-regex': '.{0,500}',
-        'description': 'Brief summary of your review.',
+        'description': 'Brief (500 char max) summary of your review.',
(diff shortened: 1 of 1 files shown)


venues/ICML.cc/2017/WHI/python/config.py
```diff
@@ -100,12 +100,12 @@
     'title': {
         'order': 1,
         'value-regex': '.{0,500}',
-        'description': 'Brief summary of your review.',
+        'description': 'Brief (500 char max) summary of your review.',
```
mspector: looks good, but I think this is wrong. Should be "(5000 char max)", right?

pmandler: The title is 500 char, the review was 5k.

mspector: oh I'm sorry, you're right, I'm not sure why I got that confused

PR #431, opened 2017-07-03
Feature/whi requests notebook

mspector:
We don't normally commit .ipynb files, but I'm committing this one because I think we need a better way of keeping track of program chair requests and sharing the solutions between ourselves.

I propose to add a "requests.ipynb" file to each venue's /python directory. It will contain the solutions to pop-up requests made by program chairs, and any notes or comments related to those requests.

This one is for WHI, and includes a solution to add revisions to papers. I also created a .py script that does the same thing.


mspector:
@mzarozinski since @pmandler is out for a while, could you take a quick look at this?

PR #433, opened 2017-07-05
Enhancement/dblp upload

mzarozinski:


venues/DBLP.org/python/dblp_data.py
```diff
@@ -0,0 +1,3 @@
+BASE = 'DBLP.org'
+GROUP = BASE + '/upload'
```
mspector: After looking through this, I'm not sure that we need this "DBLP.org/upload" group after all. Does it serve some purpose that the DBLP.org group can't?

mzarozinski: updated in latest push.

PR #434, opened 2017-07-07
Fix/remove templates utils

mspector:
removed the "utils" package, which contained the modules utils.py and templates.py

refactored all scripts using the utils package

from now on, we should define invitation parameters directly in the config file. the admin folder already has templates for these, in the config.py template


mspector:
@mzarozinski , do you mind taking a quick look at this PR?

I think pam is probably the person that this is most relevant for, but I don't want it to stay open for too long, since it touches a lot of files

mzarozinski:
caused a merge conflict with https://github.com/iesl/openreview-scripts/pull/435

PR #442, opened 2017-08-04
bug fixes in superuser init script

mspector:
URL parameter was missing, causing an error


pmandler:
Bugs fixed.  Before you close this out, mind make a few usability changes?
1) Move the "Enter the duedate" right after "Enter the human-readable due date string that will ...."
2) Catch exceptions (specifically date due) and allow user to re-enter so don't lose all the info already entered.

mspector:
@pmandler I made the requested changes, but I think the long term solution is to just move away from this text-based input. I may make that change at some point.

PR #444, opened 2017-08-08
Feature/desemweb assignment updates

mspector:


pmandler:
Please add a comment saying the expected file format - it's not hard to figure out, but still.  
Can it handle less than 3 reviewers?  Maybe that's a given for this conference.

mspector:
@pmandler added a help text to the file argument.

It can't handle less than three reviewers at the moment. I don't think it's worth changing

PR #447, opened 2017-08-14
new schema for matching data

mspector:
bug fixes in superuser init script

moved matching script into the admin directory, for corresponding changes in openreview-matching

rework matching API

remove search-based feature from uai_features.py

add standard metadata file, which will eventually become more generic

update some variable names


venues/auai.org/UAI/2017/python/basic-assignment-note.py
```diff
@@ -0,0 +1,55 @@
...
+    print "clearing assignment notes..."
+    configs = client.get_notes(invitation=config.ASSIGNMENT)
+    for n in configs:
+        client.delete_note(n)
+
+
+clear_configs()
+
+matching_configuration = {
+    "assignment_label": 'reviewers',
```
zbialecki: Could we rename this to just label?

venues/auai.org/UAI/2017/python/config.py
```diff
@@ -8,6 +8,8 @@
 PC = "auai.org/UAI/2017/Program_Committee"
 CONFERENCE = "auai.org/UAI/2017"
 SUBMISSION = "auai.org/UAI/2017/-/blind-submission"
+METADATA = "auai.org/UAI/2017/-/Paper_Metadata"
+ASSIGNMENT = 'auai.org/UAI/2017/-/Matching_Assignments'
```
zbialecki: I feel like Matching_Assignments is a little redundant. Could we make it Reviewer_Assignments?
Also, this string should probably use double quotes like the others.

zbialecki: I also like that because then we could have a different invitation for Areachair_Assignments

mspector: I'm not attached to the name Matching_Assignments, but this invitation is capable of handling assignments from both reviewers and areachairs. Each note has all the information it needs (including the ID of the group that it matches), so having separate invitations would be redundant.

That being said, exactly how to organize it would be up to the conference administrator, and they could choose to do something like this

venues/auai.org/UAI/2017/python/config.py
```diff
@@ -8,6 +8,8 @@
 PC = "auai.org/UAI/2017/Program_Committee"
 CONFERENCE = "auai.org/UAI/2017"
 SUBMISSION = "auai.org/UAI/2017/-/blind-submission"
+METADATA = "auai.org/UAI/2017/-/Paper_Metadata"
+ASSIGNMENT = 'auai.org/UAI/2017/-/Paper_Assignments'
```
mspector: @zbialecki  How do you feel about "Paper_Assignments"?

zbialecki: I like that

PR #455, opened 2017-09-11
Fix/iclr demo webfield

mspector:
This is my initial take on the tabbed webfield for ICLR 2018

conferenceWebfield_tabs.html is running on the dev instance: http://dev.openreview.net/group?id=ICLR.cc/2018/Conference


zbialecki:
I'm confused - do you need my input on this or are you starting a new branch?

zbialecki:
One thing to note is that new-style webfields should not be html files, just pure js files (i.e. no script tags). If openreview sees the script tags in the webfield content it will assume it's an old-style conference.

mspector:
Sorry, I created the pull request, but realized that really I just needed to start a branch and share it with you.

And thanks for letting me know about the new style of webfield - I think you told me this before but I had forgotten. I will change it now

PR #456, opened 2017-09-21
Nips 2017

pmandler:
Forgot to name it feature.
This is the code for the two new NIPS 2017 Workshops, MLITS and Autodiff.


venues/NIPS.cc/2017/Workshop/Autodiff/params.json
```diff
@@ -0,0 +1,14 @@
+{
```
mspector: Do you need this file and the file /data/NIPS2017_autodiff.json ? they do look different but they look like they're achieving similar things...

venues/NIPS.cc/2017/Workshop/MLITS/params.json
```diff
@@ -0,0 +1,14 @@
+{
```
mspector: same here. what's the difference between these two json files?

venues/NIPS.cc/2017/Workshop/Autodiff/params.json
pmandler: They are the files used by admin/superuser_init.py to create the conferences.  So yes they are very similar, but different titles and due dates.

mspector: I still don't fully understand why we need both of them. The `conference_title`s look the same to me... am I missing something?

venues/NIPS.cc/2017/Workshop/MLITS/params.json
pmandler: One says 
"MLITS Workshop" - "Machine Learning for Intelligent Transportation Systems", and is due on Oct 12 the other says 
"Autodiff Workshop" - "The future of gradient-based machine learning software and techniques" and is due on Oct 14.

mspector: Yes, I understand that there are two different conferences, but there are four total files, two of which seem redundant:

venues/NIPS.cc/2017/Workshop/MLITS/params.json seems redundant with venues/NIPS.cc/2017/Workshop/MLITS/data/NIPS2017_MLITS.json,

and 

venues/NIPS.cc/2017/Workshop/Autodiff/params.json seems redundant with
venues/NIPS.cc/2017/Workshop/Autodiff/data/NIPS2017_autodiff.json

melisabok:
What about this? should we merge it? is this code already in live?

pmandler:
Yes it should be merged.  And yes it is already live (oops). 
I tried to remove the redundant files NIPS2017_MLITS.json and the autodiff version. They are gone on my copy, but I’m not sure I got them out of this check in.

Pam
___________________
Pam Mandler
Center for Data Science

PR #457, opened 2017-09-22
Feature/iclr demo

mspector:


mspector:
@melisabok I can walk you through these changes if you'd like

mspector:
@zbialecki I think you might still be working on this branch, is that right? unless you object, I think I would like to merge this anyway, since the branch is getting so full of changes, and it does reflect what is currently on the dev instance

zbialecki:
I was working on the fix/iclr-demo-webfield branch, I'm a little confused as to how my commits ended up on this branch

mspector:
Sorry, my mistake. This branch is the result of my changes to the demo, and I merged fix/iclr-demo-webfield into it at some point so that I could keep track of everything that is currently deployed to the dev instance.

Is this an OK way to do things? when it comes time to merge your new changes from fix/iclr-demo-webfield, I think it should be fine

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
```diff
@@ -35,7 +51,8 @@ function() {
       readers: ['everyone'],
       content: {
         authors: ['Anonymous'],
-        authorids: ['Anonymous']
+        authorids: ['Anonymous'],
```
mspector: can set these authorids to be the authorgroup without having to overwrite it below

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
```diff
@@ -35,7 +51,8 @@ function() {
...
+        _bibtex: getBibtex(note)
       }
     }
```
mspector: remember to send a message linking the blinded note, not the original note

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
```diff
@@ -165,26 +189,104 @@ function() {
           }
         }
 
+        var publicCommentInvitation = {
```
mspector: write a new "lightweight" version of this, reserved for the last day of the submission period

zbialecki:
@mspector I think doing it this way will cause serious merge conflicts when I merge my branch in, because it further modifies the same webfield code. I'm not sure that we can totally avoid this at this point but merging in the lastest code from my branch should reduce the conflicts I have to deal with.

mspector:
can we merge fix/iclr-demo-webfield before we merge this one? then I can deal with the conflicts

zbialecki:
Yeah I think that would be good.

melisabok:
Can we merge this branch? then @pmandler  can work on this version of the process function.

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
```diff
@@ -5,8 +5,24 @@ function() {
     var CONF = 'ICLR.cc/2018/Conference';
     var PROGRAM_CHAIRS = CONF + '/Program_Chairs';
     var AREA_CHAIRS = CONF + '/Area_Chairs';
-    var BLIND_SUBMISSION = CONF + '/-/Blind_Submission';
+    var REVIEWERS = CONF + '/Reviewers';
     var AUTHORS = CONF + '/Authors';
+    var REVIEWERS_PLUS = REVIEWERS + '_and_Higher';
```
melisabok: What does it mean and_Higher? I'm not sure that this is the way to go.

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
```diff
@@ -165,26 +189,104 @@ function() {
...
           or3client.or3request(or3client.grpUrl, authorGroup, 'POST', token),
-          or3client.or3request(or3client.inviteUrl, withdrawPaperInvitation, 'POST', token),
           or3client.or3request(or3client.grpUrl, reviewerGroup, 'POST', token),
           or3client.or3request(or3client.grpUrl, areachairGroup, 'POST', token),
           or3client.or3request(or3client.grpUrl, anonReviewer1Group, 'POST', token),
           or3client.or3request(or3client.grpUrl, anonReviewer2Group, 'POST', token),
           or3client.or3request(or3client.grpUrl, anonReviewer3Group, 'POST', token),
           or3client.or3request(or3client.grpUrl, anonReviewer4Group, 'POST', token),
           or3client.or3request(or3client.grpUrl, anonReviewer5Group, 'POST', token),
+          or3client.or3request(or3client.inviteUrl, withdrawPaperInvitation, 'POST', token),
```
melisabok: I realized that we can get the 'deleted' notes if that is the issue of displaying the withdrawn paper in the ICLR home page. We wouldn't need this invitation and we can user the current deletion functionality.

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
mspector: "reviewers plus" is a group consisting of reviewers, area chairs, and program chairs.
"area chairs plus" is just area chairs and program chairs.

we do it this way because the program chairs want people to be able to select a single "readers" option, and they want the options to be organized this way.

melisabok: I see, but the Reviewers group is not created by paper number, right? Same as plus groups. We only need to create once per all the conference. It shouldn't be in the process function, may be in the superuser-init?

mspector: that's right, it's not dependent on paper number - but it's not being created in this process function. It's just a string variable that gets used later, in the "readers" field of the comment invitations. I found it easier to read this way.

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
mspector: Actually, there's a potential issue here: the author group needs to get the number of the blinded note, which only happens after it gets posted. We could use the un-blinded note's number, but in UAI, I noticed that it was possible for the blinded and un-blinded note numbers to get off-sync from each other. I don't know if it's worth the risk of that.

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
mspector: ok, that makes sense. My only worry with this solution is that it may be confusing for the user. They may expect the "delete" functionality to delete it and hide it. As opposed to "withdraw", which doesn't necessarily imply that the note will be hidden

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
melisabok: Yes, the number should be the number of the blinded note, that is how UAI works.

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
melisabok: Ohhh, I'm sorry, discard my comment!

venues/ICLR.cc/2018/Conference/process/submissionProcess.js
melisabok: Look at the UAI process function: https://github.com/iesl/openreview-scripts/blob/master/venues/auai.org/UAI/2017/process/submissionProcess.js#L60

We create the blinded note and then the author group using the blinded note paper number

melisabok:
Given that this code is already deployed in the dev instance I'm going to merge it. So we keep synced dev with master code.

PR #458, opened 2017-09-25
Fix/ICLR Demo Webfield

zbialecki:
Add a few features to the ICLR group webfield and fix a few bugs with the Add Bids invitation webfield


zbialecki:
This PR https://github.com/iesl/openreview/pull/787 to be merged before it can be deployed

PR #463, opened 2017-10-03
add originals tab and instructions

mspector:


mspector:
@melisabok @zbialecki 

this is my attempt to make our anonymity mechanism more clear to ICLR authors. What do you guys think?

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -95,7 +99,13 @@ function renderConferenceHeader() {
     location: "Vancouver Convention Center, Vancouver, BC, Canada",
     date: "April 30 - May 3, 2018",
     website: "http://www.iclr.cc",
-    instructions: null,  // Add any custom instructions here. Accepts HTML
+    instructions: `<b>Important Information about Anonymity:\n</b>
```
mspector: @melisabok @zbialecki I'm having trouble getting this to format correctly. any idea what I should be doing differently?

melisabok: I think \n is not html code.

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -95,7 +99,14 @@ function renderConferenceHeader() {
...
+    Please read these instructions carefully.\n
+    When you post a submission to ICLR 2018, please provide your real name and email address in the submission form below.
+    An anonymous copy of your paper will appear in the "All Submitted Papers" tab below.
+    An <em>original</em> version of your paper will be available in the "My Original Papers" tab.
+    You can also access the original version of your paper by clicking the "Modifiable Original"
```
melisabok: Also through the main Task view.

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -125,6 +136,10 @@ function renderConferenceTabs() {
       id: 'my-submitted-papers',
     },
     {
+      heading: 'My Original Papers',
```
melisabok: So, what is the difference between 'My Submitted Papers' and 'My Original Papers'?  One displays the blinded notes other the original notes?

For me is enough showing only one tab(Submitted Papers) and telling to the user that they can access to the original through the 'Modifiable Original' link or Task view. (my opinion, I may be wrong)

mspector: yes, one shows blinded, the other shows originals.

yeah, I'm not sure either. but my thought is that showing both these tabs next to each other (with these instructions right above) will prompt the user to understand what's really happening sooner in the submission process (i.e. as soon as they submit a paper, they'll see that there is both an original and a copy).

it also means that the user doesn't have to go to the tasks page as often, which I think is a good thing (since we want to redesign it soon). 

but yes, I agree that it's not obviously the right solution. maybe zach has an opinion. alternatively we could show it to the program chairs on the dev site and see what they think

melisabok: Both tabs are going to show the same data with a slight difference: the author names, and in most of the cases is going to be one only paper. I think is redundant to show both and also I wouldn't like to display them separately, because they are the same thing, they represent the same paper, they are linked.

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
zbialecki: Also, for all JS code running in the browser, we shouldn't use ES6 features like template strings. I have been making sure all the code runs on browsers back to IE11, you can refer to caniuse.com if you have questions about a specific API.

One other note, to make text bold use the `<strong>` tag not `<b>`

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
zbialecki: I agree with Melisa that it seems redundant and a little confusing at first. Can we do what we do in other places and have some indication that the author name is only visible to you? ("privately revealed to you").

The other alternative I can see is removing the My Submissions tab and just keeping My Originals

mspector: I don't know if adding that text would help the problem I'm trying to solve here, which is that users don't have a good idea about how the original/copy system works. But maybe the instructional text is enough.

If we're going to keep one of these tabs, I do think it needs to be the "My Submissions" tab - I think easy access to the discussion of the papers (which is only present in the blinded notes) is more important than easy access to the originals.

I'll take out the my originals tab and fix the text.

melisabok: I agree with that, also you can add a tooltip in the 'Modifiable Original' button with some help text. @zbialecki can help you to do that.

zbialecki: Adding a tooltip to a button is easy - you just have to add a little markup to the html and call a JS function after the button is added to the page.
https://getbootstrap.com/docs/3.3/javascript/#tooltips

mspector:
@melisabok @zbialecki I removed the originals tab and changed the wording / formatting of the instructional text. Do you guys mind checking it over? let me know if you think it should be different.

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -95,7 +95,18 @@ function renderConferenceHeader() {
     location: "Vancouver Convention Center, Vancouver, BC, Canada",
     date: "April 30 - May 3, 2018",
     website: "http://www.iclr.cc",
-    instructions: null,  // Add any custom instructions here. Accepts HTML
+    instructions: '<p><strong>Important Information about Anonymity:</strong><br />\
```
zbialecki: HTML5 does not use self-closing tags, just use `<br>` in this case.

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -95,7 +95,18 @@ function renderConferenceHeader() {
...
+    When you post a submission to ICLR 2018, please provide your real name and email address in the submission form below.\
+    An anonymous copy of your paper will appear in the "All Submitted Papers" tab, and will be visible to the public. \
+    The <em>original</em> version of your paper will be private, and will contain your real name(s); \
+    originals can be found in your OpenReview <a href=/tasks>Tasks page</a>.\
+    You can also access the original version of your paper by clicking the "Modifiable Original" \
+    link in the discussion forum page of your paper. <br /><br /> \
```
zbialecki: Instead of using 2 brs to make a new paragraph it's better if you use `<p>` tags. Just close the previous p tag and open a new one for the new paragraph.

mspector:
oops, I should have included that. here it is:

iclr18-text

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -95,7 +95,18 @@ function renderConferenceHeader() {
     location: "Vancouver Convention Center, Vancouver, BC, Canada",
     date: "April 30 - May 3, 2018",
     website: "http://www.iclr.cc",
-    instructions: null,  // Add any custom instructions here. Accepts HTML
+    instructions: '<p><strong>Important Information about Anonymity:</strong><br />\
+    When you post a submission to ICLR 2018, please provide your real name and email address in the submission form below.\
+    An anonymous copy of your paper will appear in the "All Submitted Papers" tab, and will be visible to the public. \
+    The <em>original</em> version of your paper will be private, and will contain your real name(s); \
+    originals can be found in your OpenReview <a href=/tasks>Tasks page</a>.\
```
zbialecki: HTML attributes should always use quotes. This should be `<a href="/tasks">`

venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -95,7 +95,18 @@ function renderConferenceHeader() {
...
+    You can also access the original version of your paper by clicking the "Modifiable Original" \
+    link in the discussion forum page of your paper. <br /><br /> \
+    <strong>Posting Revisions to Submissions:</strong><br />\
+    To post a revision to your paper, navigate to the original version, and click on the "Add Revision" button if available. \
+    Revisions are not allowed during the formal review process.\
+    Revisions on originals propagate all changes to anonymous copies, while maintaining anonymity. <br /><br /> \
+    Please contact the OpenReview support team at <a href=mailto:info@openreview.net>info@openreview.net</a> with any questions or concerns.</p>',  // Add any custom instructions here. Accepts HTML
```
zbialecki: Attribute needs quotes

melisabok:
This is deployed in the dev site but I found that the Deadline lost the format.

screen shot 2017-10-04 at 1 54 20 pm

@zbialecki do you have an idea what is the problem here?

zbialecki:
In the inspector it looks like the HTML is messed up - tags aren't closed properly. It's weird because I don't see problems with Michael's HTML string.

zbialecki:
Interesting, you aren't allowed to nest p tags https://stackoverflow.com/questions/12015804/nesting-p-wont-work-while-nesting-div-will
Let me make a change to the venue header template.

PR #464, opened 2017-10-03
Feature/iclr email template

mspector:
this is pretty similar to last year's script.


melisabok:
is the same file than the previous conference? can we put the file in a common place instead of duplicate it?

mspector:
good idea. It's not exactly the same but you're right, it should be in a common place.

I suggest putting it in openreview-scripts/samples. is that ok?

PR #465, opened 2017-10-03
Fix/update recruit reviewers

mspector:
allows the script to accept a list of emails. this is useful for sending recruitment messages to one or two extra people that we missed the first time around.


melisabok:
what about this PR? @pmandler ?

PR #467, opened 2017-10-04
Fix bug where submitting a new paper would cause all the papers to be rendered twice

zbialecki:


zbialecki:
Needs https://github.com/iesl/openreview/pull/797 to work properly

PR #468, opened 2017-10-04
fix bibtex year

mspector:
I have a jupyter notebook script that will correct the currently mislabeled bibtex records


melisabok:
Good catch!

PR #469, opened 2017-10-05
Fix/nips autodiff due dates

pmandler:


mspector:
there are lots of changes to ICLR files that I don't think we want. It looks like you committed these changes on a stale branch

PR #473, opened 2017-10-11
Refresh conference Authors and paper Authors groups using latest pape…

pmandler:
…r revision.

### DIFF
--- a/samples/python/refresh-authors.py
+++ b/samples/python/refresh-authors.py
@@ -1,0 +1,91 @@
+#!/usr/bin/python
+
+###############################################################################
+# ex. python refresh-authors.py --conf MyConf.org/2017 --baseurl http://localhost:3000
+#       --username admin --password admin_pw
+#
+# To be run to refresh author groups in case of revisions and/or withdrawals
+# Expects config to define BLIND_SUBMISSION, SUBMISSION and AUTHORS as well
+# as an Authors group for each SUBMISSION paper
+###############################################################################
+
+## Import statements
+import argparse
+import sys
+from openreview import *
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('-c','--conf', required=True, help = "the full path of the conference group ex. MyConf.org/2017")
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+## Initialize the client library with username and password
+client = Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+## check conference directory exists
+base_path = "../../venues/"+args.conf
+config_path = base_path+"/python/"
+if os.path.isfile(config_path+"config.py") is False:
+    print "Cannot locate config.py in:"+config_path
+    sys.exit()
+## load conference specific data
+sys.path.insert(0, config_path)
+import config
+
+### update authors lists
+all_authors = []
+submissions = client.get_notes(invitation=config.SUBMISSION)
+blinded = client.get_notes(invitation=config.BLIND_SUBMISSION)
+for blind_note in blinded:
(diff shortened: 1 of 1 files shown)


samples/python/refresh-authors.py
```diff
@@ -0,0 +1,91 @@
...
+all_authors = []
+submissions = client.get_notes(invitation=config.SUBMISSION)
+blinded = client.get_notes(invitation=config.BLIND_SUBMISSION)
+for blind_note in blinded:
+    #find associated note in Submissions to get true authors
+    parent_forum = blind_note.original
+    note_authors = []
+    found_profile = False
+    #for parent_notes = submissions[].forum == note.original
+    for note in submissions:
```
mspector: instead of looping through all submissions and breaking when you find the right one, you could instead create a dictionary of forum IDs to submissions, and then index into it using the blind_note.original value.

so line 41 could be something like:

`submissions_by_forum = {n.forum: n for n in client.get_notes(invitation=config.SUBMISSION)`

Then, in this loop, you could look up the note by doing:

`note = submissions_by_forum[blind_note.original]`

samples/python/refresh-authors.py
```diff
@@ -0,0 +1,91 @@
...
+                email = email.lower()
+                try:
+                    profile = client.get_profile(email)
+                    note_authors.append(profile.id)
+                    found_profile = True
+                    #print "found profile "+email
+                except openreview.OpenReviewException as e:
+                    # no ~id, use the email
+                    note_authors.append(email)
+                    #print "no profile "+email
```
mspector: I might actually keep some kind of print statement here when an exception gets thrown... something as simple as `print "no profile found: ", e`. The reason I think it's worthwhile is because it's possible for the exception to be something other than the a "not found" error.

either that, or you could check the contents of the error to make sure that it's actually the error you think. If it's not, you can call `raise(e)` and it will throw the error as normal.

mspector:
looks great! I made a few minor inline comments.

one general comment: I'm not sure if this script really belongs in the /samples directory. In my opinion, the /samples directory should be a place for us to store sample code snippets that demonstrate the API, not robust scripts to be executed as they are. This script reads like a one that is supposed to be conference-agnostic (checking for the existence of the directories, for example). 

@melisabok @pmandler what do you think about this? I think robust, conference-agnostic scripts that can be run just as they are should go in the /admin directory.

samples/python/refresh-authors.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # if we don't have any profiles for the authorids,
+            # add the ID of the person who submitted the paper.
+            if found_profile == False:
+                note_authors.append(note.signatures[0])
+
+            # update paper author group
+            author_group = client.get_group(config.CONF+"/Paper"+num_str+"/Authors")
```
melisabok: The author group should use the paper number of the blinded note instead of the original note, am I right @mspector ?

mspector: good catch! yes, that is correct

samples/python/refresh-authors.py
melisabok: For any error that you could get from the get_profile call, it should be the email the default value, so I think it is ok.

samples/python/refresh-authors.py
melisabok: Yes, using dictionaries is cheaper than loop the all the notes every time.

melisabok:
Ok, let's move it to the admin folder. And this script should support conferences with and without double-blind configuration, right?

melisabok:
May be instead of 'Admin' we can call the folder 'Tools'? or 'Utils'?

mspector:
I like /tools or /utils. I'm moving things around in my refactor of the conference templates, so maybe we can just merge this in the /samples folder, and I'll reorganize it in my PR. does that sound ok?

PR #477, opened 2017-10-18
Script to change uppercase emails to lowercase.

pmandler:
I don’t believe there are no actual profiles that refer to uppercase email addresses.  Is there a check somewhere in the code that makes this the case in which case I don’t have to worry about testing for this at all?
For now I’m just printing an error if uppercase profiles or groups are found.

### DIFF
--- a/utils/group-emails-to-lowercase.py
+++ b/utils/group-emails-to-lowercase.py
@@ -1,0 +1,95 @@
+#!/usr/bin/python
+
+###############################################################################
+# ex. python group-emails-to-lowercase.py -group MyConf.org/2017/Reviewers --baseurl http://localhost:3000
+#       --username admin --password admin_pw
+#
+# Sets all members that are emails to lowercase.  Notifies user if there are profiles or groups using the
+# uppercase letters in the
+###############################################################################
+
+## Import statements
+import argparse
+import sys
+from openreview import *
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('-g','--group', required=True, help = "the full path of the group ex. MyConf.org/2017")
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+## Initialize the client library with username and password
+client = Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+def get_profile(value):
+    profile = None
+    try:
+        profile = client.get_profile(value)
+    except openreview.OpenReviewException as e:
+        # throw an error if it is something other than "not found"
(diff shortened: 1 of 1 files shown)


melisabok:
Do you mean only for the reviewers group, right? I bet there are profiles with uppercase emails in the system. 

It is good to print them out.

pmandler:
It was made for the reviewers group, but it could be used on other groups if you want. It looks for the ‘@‘ so it should only effect emails.

Pam
___________________
Pam Mandler
Center for Data Science

utils/group-emails-to-lowercase.py
```diff
@@ -0,0 +1,95 @@
...
+        profile = client.get_profile(value)
+    except openreview.OpenReviewException as e:
+        # throw an error if it is something other than "not found"
+        if e[0][0] != 'Profile not found':
+            print "OpenReviewException: {0}".format(e)
+            raise e
+    return profile
+
+def get_group(value):
+    group = None
```
melisabok: May be we should move this logic to the library, if the profile or group doesn't exist then return None instead of raise an exception

utils/group-emails-to-lowercase.py
```diff
@@ -0,0 +1,95 @@
...
+    has_upper = False
+    for character in value:
+        if character.isupper():
+            has_upper = True
+    return has_upper
+
+# locate emails w/ uppercase letters
+#   replace with lowercase email
+def upper_to_lower(group_name):
+    target_group = client.get_group(group_name)
```
melisabok: here you are not using your internal function, why? what happens if the group_name doesn't exist?

utils/group-emails-to-lowercase.py
```diff
@@ -0,0 +1,95 @@
...
+
+# print uppercase profiles and groups found
+def find_upper_email_references(upper_emails):
+    ## try to get profile via email
+    ## if get one, check that lower exists and set to that group instead
+    for email in upper_emails:
+        profile = get_profile(email)
+        if profile != None:
+            print "ERROR: Uppercase profile for {0} with id {1}".format(email, profile.id)
+        group = get_group(email)
```
melisabok: This is unnecessary becuase the uppercase email always exists, it was a member of the reviewer group. Also the lowercase email is created when you call 'add_members_to_group'.

What we have to do is:
- if the uppercase email is part of the profile, replace it with the lowercase email. 
- replace all the member relationships with the lowercase group.
- replace all the signatures, readers, writers or tauthor where can appear the uppercase email with the lowercase email. (I guess this never happens but I'm not sure).
- remove uppercase group.

mspector:
@pmandler @melisabok , is this ready for merging?

PR #483, opened 2017-10-23
Fix/ICLR homepage webfield

zbialecki:
Fixes
- All Submitted Papers search displays original and blinded notes
- My Submitted Papers shows comments as well as papers


mspector:
Looks good!

Sorry if I missed this (I feel like I haven't been tracking these issues that closely), but do these changes need to be made in the "non-alt" version of the webfield too?

melisabok:
I think we should keep working on the Alt version and deploy it in live. And go back to the other version later. Having two version is inefficient.

zbialecki:
Yeah, agreed. Before you merge this Michael there's actually another bug I wanted to fix. I emailed you guys about it.

PR #487, opened 2017-10-25
Feature/refresh bibtex script

mspector:


venues/ICLR.cc/2018/Conference/python/refresh-bibtex.py
```diff
@@ -0,0 +1,41 @@
...
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+submissions = client.get_notes(invitation=config.BLIND_SUBMISSION)
+
+for n in submissions:
+    # a submission potentially has many references, but we want to change only
+    # the "prime" reference, which has an ID equal to its referent.
+    refs = client.get_revisions(referent=n.id)
```
melisabok: Why don't you add a new revision with the new value?

BTW what you are doing is exactly to update note.

mspector: Good point!

mspector: It just occurred to me that we might not want the bibtex changes to appear in the revision history. what do you think @melisabok ?

melisabok: But _bibtex is hidden in the UI, right?

Ok, you can do an update then.

melisabok: I remember that you said that you were going to do this in the Add/Revision process function, what happened?

mspector: ah, maybe you are right, that _bibtex is hidden.

well, I didn't do it because I originally thought it wasn't necessary. But it would be better if we can do it. So I will implement that now.

melisabok:
should I merge this?

mspector:
@melisabok  yes I think it can be merged

melisabok:
@mspector I added the forum url and refactored the code, post the first reference is the same that updating the note.

melisabok:
I reverted the refactor, I'll leave the PR is open until we can fix the OpenReviewPy library

melisabok:
I fixed this script and depends on this PR: https://github.com/iesl/openreview-py/pull/63

mspector:
@melisabok I approve! I just tried to merge it but I can't, since I created it initially

mspector:
oops, never mind, I actually can (and did)

melisabok:
Cool, I didn’t run the script in live

PR #493, opened 2017-10-28
add deduplication script

mspector:


venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
```diff
@@ -0,0 +1,94 @@
...
+
+# iterate through the duplicates
+for paperhash, notes in duplicates_by_paperhash.iteritems():
+
+    # for every paperhash, sort the notes by tcdate.
+    # the first note should be the one posted last.
+    latest_note = sorted(notes, key=lambda n: n.tcdate, reverse=True)[0]
+    assert latest_note.tcdate == max([n.tcdate for n in notes])
+
+    print "keeping ",latest_note.id
```
melisabok: I thought you wanted to keep the first submission?

venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
```diff
@@ -0,0 +1,94 @@
...
+
+            try:
+                deleted_n = client.post_note(n)
+                print "--> deleting original ", deleted_n.id
+            except openreview.OpenReviewException as e:
+                print 'ERROR on {0}:{1} '.format(n.id, e)
+
+
+            try:
+                blind_ref = blindrefs_by_original[n.id]
```
melisabok: why are you deleting the reference and not the blinded note?

venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
```diff
@@ -0,0 +1,94 @@
...
+                blind_ref.ddate = 1509138000000
+                deleted_ref = client.post_note(blind_ref)
```
melisabok: I'm not sure that deleting the reference will delete the note

venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
mspector: No, I think the last one is better. Sometimes people submitted twice, thinking the second would overwrite the first

venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
mspector: I thought this was the proper way to do it, but I'll check again, maybe you are right

venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
melisabok: I checked the backed and we you update a reference with the same id as the note then we do an update note, so I'm not sure what is the content of the blinded note after this, could you check?

venues/ICLR.cc/2018/Conference/python/deduplicate-submissions.py
melisabok: I think you have to update the blinded note instead of the reference.

PR #494, opened 2017-10-28
add script to fix blind submissions

mspector:


venues/ICLR.cc/2018/Conference/python/fix-blinded-submissions.py
```diff
@@ -0,0 +1,74 @@
...
+    blind_submission_params = {
+        'original': original_note.id,
+        'invitation': config.BLIND_SUBMISSION,
+        'forum': None,
+        'signatures': [config.CONF],
+        'writers': [config.CONF],
+        'readers': ['everyone'],
+        'content': {
+            'authors': ['Anonymous'],
+            'authorids': ['ICLR.cc/2018/Conference/-/Paper{0}/Authors'.format(original_note.number)],
```
melisabok: I see that we are using the original note number here, but it is not what the submission process function does:

https://github.com/iesl/openreview-scripts/blob/master/venues/ICLR.cc/2018/Conference/process/submissionProcess.js#L108

melisabok: numbers of different invitations(original and blind) are not synced!!!

mspector: I see. I can change this if you'd like, but I've been running scripts to make sure that the blind note and original notes have the same numbers

melisabok: ok, you already run this script and it worked so I'll merge it.

mspector: the reason the blind/original are not synced is because there are two withdrawn notes after the deadline. withdrawing after the deadline deletes the blind, keeps the original, and creates a "withdrawn submission"

PR #497, opened 2017-10-30
add author groups script

mspector:
this script regenerates the missing paper and author groups (which didn't get created, due to process function timeouts)

### DIFF
--- a/venues/ICLR.cc/2018/Conference/python/regen-paperauthor-groups.py
+++ b/venues/ICLR.cc/2018/Conference/python/regen-paperauthor-groups.py
@@ -1,0 +1,41 @@
+import openreview
+import config
+import csv
+import argparse
+
+## Handle the arguments
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--username')
+parser.add_argument('--password')
+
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2018/Conference/python/regen-paperauthor-groups.py
```diff
@@ -0,0 +1,41 @@
...
+parser.add_argument('--username')
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print client.baseurl
+
+# repair author groups
+submissions = client.get_notes(invitation=config.SUBMISSION)
```
melisabok: shouldn't we use the blind submission invitation number to create the groups?

mspector: well, we need to get the note's signatures and authorids to populate the authorgroup. the only risk is that the blind note's numbers are not the same as the original note's numbers, but I have an arango script that makes sure that they're aligned

melisabok: well, they shouldn't be aligned.

PR #500, opened 2017-10-30
Fix/add bid webfield

zbialecki:
Updates Add Bid webfield to use the Webfield module. It does not enable pagination though, because we need to have all note before we can sort them by bid.


zbialecki:
Requires https://github.com/iesl/openreview/pull/828

zbialecki:
Should be much faster now. Tabs are only rendered in full on the initial page load.

PR #502, opened 2017-11-02
correct authorgroups script

mspector:
this script corrects the authorgroup numbers

### DIFF
--- a/venues/ICLR.cc/2018/Conference/python/correct-authorgroups.py
+++ b/venues/ICLR.cc/2018/Conference/python/correct-authorgroups.py
@@ -1,0 +1,80 @@
+import openreview
+import config
+import csv
+import argparse
+from collections import defaultdict
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print client.baseurl
+
+# get all the authorgroups.
+# where does 1176 come from?
+# 1176 is the value on the counter for ICLR.cc/2018/Conference/-/Submission
+authorgroups = []
+print "finding valid authorgroups ..."
+for i in range(1, 1176):
+    groupid = 'ICLR.cc/2018/Conference/Paper{0}/Authors'.format(i)
+    try:
+        authorgroup = client.get_group(groupid)
+        authorgroups.append(authorgroup)
+    except:
+        pass
+
+
+# strip all the author groups of their members
+print "remove all members from all authorgroups ..."
+for authorgroup in authorgroups:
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2018/Conference/python/correct-authorgroups.py
```diff
@@ -0,0 +1,80 @@
...
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print client.baseurl
+
+# get all the authorgroups.
+# where does 1176 come from?
+# 1176 is the value on the counter for ICLR.cc/2018/Conference/-/Submission
+authorgroups = []
+print "finding valid authorgroups ..."
+for i in range(1, 1176):
+    groupid = 'ICLR.cc/2018/Conference/Paper{0}/Authors'.format(i)
```
melisabok: why don't you use this call to get all the existent groups?

https://openreview.net/groups?id=ICLR.cc/2018/Conference/Paper.*/Authors

venues/ICLR.cc/2018/Conference/python/correct-authorgroups.py
```diff
@@ -0,0 +1,80 @@
...
+print "remove all members from all authorgroups ..."
+for authorgroup in authorgroups:
+    client.remove_members_from_group(authorgroup, authorgroup.members)
+
+
+# iterate through all the blind submissions and set the authorids to the right number authorgroup
+print "iterating through blind submissions, updating them ..."
+blind_submissions = client.get_notes(invitation=config.BLIND_SUBMISSION)
+for n in blind_submissions:
+    n.content['authorids'] = ['ICLR.cc/2018/Conference/Paper{0}/Authors'.format(n.number)]
```
melisabok: before doing this I would check if the authorids fields has a different value

venues/ICLR.cc/2018/Conference/python/correct-authorgroups.py
```diff
@@ -0,0 +1,80 @@
...
+        authorgroup = client.get_group(groupid)
+        authorgroups.append(authorgroup)
+    except:
+        pass
+
+
+# strip all the author groups of their members
+print "remove all members from all authorgroups ..."
+for authorgroup in authorgroups:
+    client.remove_members_from_group(authorgroup, authorgroup.members)
```
melisabok: instead of removing and adding members you can re-post the group. When you update an existing groups all the members are removed and added again based on the input body.

This is safer because there is a moment in the scripts that the groups are empty and active users can have errors.

venues/ICLR.cc/2018/Conference/python/correct-authorgroups.py
mspector: ah, i forgot about this. i can change it

mspector:
new changes rely on this PR:

https://github.com/iesl/openreview-py/pull/64

PR #503, opened 2017-11-06
update review due date and send emails to reviewers

pmandler:


venues/NIPS.cc/2017/Workshop/MLITS/python/email-reviewers.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Thank you for your help on evaluating this paper. The success of our workshop is to a large extent due to the efforts of our committee members.
+
+    Sincerely, the organizing committee,
+
+    Li Erran Li
+    Anca Dragan
+    Juan Carlos Niebles
+    Silvio Savarese""".format(paper_id)
+    response = client.send_mail(subjectline, [email_addr], message)
```
mspector: if you wanted, you could return this response object, and then print it out in your main loop below. this is sometimes useful for debugging

PR #508, opened 2017-11-09
Feature/pc reviewer assignment

mspector:
I wanted to reuse the UAI script, but it was much more complicated because it tried to detect conflicts.

I think we should leave it up to the PCs (or us) to detect conflicts before assigning.


venues/ICLR.cc/2018/Conference/python/reviewer-matching.py
```diff
@@ -98,7 +98,7 @@ def metadata(forum):
         "label": 'reviewers',
         "group": config.REVIEWERS,
         "submission": config.BLIND_SUBMISSION,
-        "exclude": [n.forum for n in withdrawn_papers],
+        "exclude": [],
```
mspector: this change is just because we changed the way that paper withdrawals work

PR #520, opened 2017-11-14
move PC scripts to programchairs directory

mspector:
so that the program chairs can more easily see which scripts are for them to run

@pmandler , you can put the scripts you're working on in this directory


melisabok:
@mspector could you update this branch please? then we can merge it

PR #521, opened 2017-11-14
add ability to output each paper with all reviewers or each reviewer …

pmandler:
…with all papers


melisabok:
Sounds good!

Given that --p and --r only works with the parameter --all. I would merge these in two options:
- all by paper
- all by reviewer

@mspector  How can I define a parameter that has only two options that are mutually exclusive.

mspector:
good question. I'm not sure if it's possible (this is something that I would like to know too). 

I'll check out the argparse library docs to see.

melisabok:
Also @pmandler can we print the help when there is no arguments set?

mspector:
Check out the second-most up voted answer to this question:

https://stackoverflow.com/questions/7869345/how-to-make-python-argparse-mutually-exclusive-group-arguments-without-prefix

It looks like that's a good way to do mutually exclusive arguments

melisabok:
This looks great, one small change. When we get the -r (reviewers), can we list the reviewers based on this group: https://openreview.net/groups?id=ICLR.cc/2018/Conference/Reviewers, otherwise we can not see the reviewers without any assignment, like victor zhong

melisabok:
thanks Pam!!

PR #525, opened 2017-11-20
Feature/author precreate

pmandler:
May want to change output to a csv file for the profiles that have different emails, but may be the same person.

### DIFF
--- a/utils/author-precreate-profile.py
+++ b/utils/author-precreate-profile.py
@@ -1,0 +1,127 @@
+#!/usr/bin/python
+
+###############################################################################
+# ex. python author-precreate-profile.py --baseurl http://localhost:3000
+#       --username admin --password admin_pw
+#
+# Find authors and precreate profiles if they are missing.
+# Possible to change out emails w/ tildeId's in future.
+###############################################################################
+
+## Import statements
+import argparse
+from openreview import *
+
+## Argument handling
+parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


utils/author-precreate-profile.py
```diff
@@ -0,0 +1,124 @@
...
+                'first': first,
+                'middle': '',
+                'last': last,
+                'username': tilde_id
+            }]
+        }
+    })
+
+    # Why doesn't  "client.post_note(profile)" work? throws OpenReviewException: [u'note does not exist']
+    response = requests.put(client.baseurl+"/user/profile", json=profile.to_json(), headers=client.headers)
```
melisabok: We need a specific call to create profiles because we are defining the id outside of the call. POST notes set automatically a string id. This is failing because POST notes receive an id in the body and thinks that you are trying to edit a note, but the note doesn't exist. 

Obs: http PUT operations is used to create object and POST to update objects. We should split the POST notes in PUT and POST depending what you want to do.

utils/author-precreate-profile.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+# for all submissions get authorids, if in form of email address, try to find associated profile
+# if profile doesn't exist, create one
+offset = 0
+limit = 200
+while True:
```
melisabok: instead of doing a 'while true' you could check the condition ```while len(notes) < limit```, so you don't to break the loop manually.

PR #528, opened 2017-11-22
Feature/rebuttal period

mspector:
these are new and updated scripts to transition ICLR from the reviewing period to the rebuttal period.

to make the transition, I will do the following:

1) reveal the existing reviews to the public and change the invitation so that future reviews will be public too:
`python visibility.py --show reviews`

2) enable revisions on all existing papers
`python invitations.py --enable Add_Revision`

3) add revision invitations to official reviews, so that we can keep track of review revisions during the rebuttal period. Also update the process function so that revisions are created when late reviews are made.
`python review-revisions.py`

### DIFF
--- a/venues/ICLR.cc/2018/Conference/process/officialReviewProcess_withRevision.js
+++ b/venues/ICLR.cc/2018/Conference/process/officialReviewProcess_withRevision.js
@@ -1,0 +1,61 @@
+//EDIT ME
+
+function(){
+    var or3client = lib.or3client;
+
+    var origNote = or3client.or3request(or3client.notesUrl+'?id='+note.forum, {}, 'GET', token);
+    var list = note.invitation.replace(/_/g,' ').split('/');
+    list.splice(list.indexOf('-',1));
+    var conference = list.join(' ');
+
+    origNote.then(function(result) {
+      var forum = result.notes[0];
+      var note_number = forum.number;
+
+      var reviewers = ['ICLR.cc/2018/Conference/Paper' + note_number + '/Reviewers'];
+      var areachairs = ['ICLR.cc/2018/Conference/Paper' + note_number + '/Area_Chair'];
+      var authors = forum.content.authorids;
+
+      var areachair_mail = {
+        "groups": areachairs,
+        "subject": "Review posted to your assigned paper: \"" + forum.content.title + "\"",
+        "message": "A submission to " + conference + ", for which you are an official area chair, has received an official review. \n\nTitle: " + note.content.title + "\n\nComment: " + note.content.review + "\n\nTo view the review, click here: " + baseUrl + "/forum?id=" + note.forum
+      };
+      var areachairMailP = or3client.or3request( or3client.mailUrl, areachair_mail, 'POST', token );
+
+      return areachairMailP;
+    })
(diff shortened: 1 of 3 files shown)


venues/ICLR.cc/2018/Conference/process/officialReviewProcess_withRevision.js
```diff
@@ -0,0 +1,61 @@
...
+    })
+    .then(result => or3client.addInvitationNoninvitee(note.invitation, note.signatures[0], token))
+    .then(result => or3client.or3request(or3client.inviteUrl+'?id='+note.invitation, {}, 'GET', token))
+    .then(result => {
+      var reviewInvitation = result.invitations[0];
+      var signatureComponents = note.signatures[0].split('/');
+      console.log('signature components: ' + signatureComponents);
+      var anonReviewerPaper = signatureComponents
+      var revisionInvitation = {
+        'id': 'ICLR.cc/2018/Conference/-/'+signatureComponents[4]+'/'+signatureComponents[3]+'/Revise_Review',
```
melisabok: We could create all the revise_review invitations in the review_revisions, so we don't need to change this process function.

melisabok:
If we want to force the reviewers to add a new revision instead of editing the current review we need to remove the anon reviewer from the writers field.

venues/ICLR.cc/2018/Conference/process/officialReviewProcess_withRevision.js
mspector: I think we need to create them in the process function, because the revision invitation needs to set its referent field to the ID of the review note, which won't exist until the review is posted

venues/ICLR.cc/2018/Conference/python/review-revisions.py
```diff
@@ -0,0 +1,60 @@
...
+    revise_review = openreview.Invitation('ICLR.cc/2018/Conference/-/{0}/{1}/Revise_Review'.format(anonreviewer, paper_num), **params)
+    return revise_review
+
+review_invitations = client.get_invitations(regex = config.CONF + '/-/Paper.*/Official_Review')
+reviews = client.get_notes(invitation = config.CONF + '/-/Paper.*/Official_Review')
+
+for inv in review_invitations:
+    forum_reviews = [r for r in reviews if r.forum == inv.reply['forum']]
+
+    inv.reply['writers'] = {'values': []}
```
mspector: the invitation reply writers gets set to empty here, which prevents reviewers from directly editing any future reviews that are posted

venues/ICLR.cc/2018/Conference/python/review-revisions.py
```diff
@@ -0,0 +1,60 @@
...
+    with open('../process/officialReviewProcess_withRevision.js') as f:
+        inv.process = f.read()
+    inv = client.post_invitation(inv)
+
+    for review in forum_reviews:
+        print "updating review: ", review.id
+        review.readers = ['everyone']
+        review.writers = []
```
mspector: this line prevents existing reviews from being directly edited

venues/ICLR.cc/2018/Conference/process/officialReviewProcess_withRevision.js
melisabok: You are right!

PR #531, opened 2017-11-28
Fix/lazy reviewers offset

mspector:


mspector:
sorry, just one more small change to avoid problems with the directory change

PR #532, opened 2017-11-28
Add AC's to output

pmandler:


melisabok:
sounds good to me

mspector:
looks good! one suggestion:

instead of formatting it like "Paper Number, AC, Reviewers" in order like that, could you do something like:

[Paper_Number, Reviewers, "(Area Chair: " + AC + ")"]

Does that make sense? I think it will make the output easier to read

melisabok:
the problem of that is that reviewers list is not fixed.

pmandler:
At first I had it paper number  AC: name, Reviewers: names, but I thought it would be unwieldy if they wanted to copy it into a spreadsheet and work with it from there.  

Pam

mspector:
ok, fair enough

PR #534, opened 2017-11-29
script to output review status

pmandler:
Combines capabilities of various 2017 get-review scripts.

### DIFF
--- a/venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
+++ b/venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
@@ -30,16 +30,19 @@ client = Client(baseurl=args.baseurl, username=args.username, password=args.pass
 
 reviews_total =0
 reviews_complete = 0
-# paper_data[paper_num]['title'] = title
-# paper_data[paper_num]['reviewers'][reviewer]['rating'/'confidence'] = rating
-# paper_data[paper_num]['AC']= Area Chair id
-# paper_data[paper_num]['reviewer_count']= # of assigned reviewers
-# paper_data[paper_num]['review_count']= # of reviews complete
+# paper_data structure
+#   paper_data[paper_num]['title'] = title
+#   paper_data[paper_num]['reviewers'][reviewer]['rating'/'confidence'] = rating
+#   paper_data[paper_num]['AC']= Area Chair id
+#   paper_data[paper_num]['meta_recommendation']= meta review recommendation
+#   paper_data[paper_num]['meta_confidence']= meta review confidence
+#   paper_data[paper_num]['reviewer_count']= # of assigned reviewers
+#   paper_data[paper_num]['review_count']= # of reviews complete
 paper_data = {}
 
 # fill in submission information
 print "Filling in submissions"
-submissions = client.get_notes(invitation=config.SUBMISSION)
+submissions = client.get_notes(invitation=config.BLIND_SUBMISSION)
 for note in submissions:
     paper_data[note.number] = {}
     paper_data[note.number]['title']= note.content['title']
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
```diff
@@ -0,0 +1,136 @@
...
+sys.path.insert(0, "./..")
+import config
+
+## Handle the arguments
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--username')
+parser.add_argument('--password')
+parser.add_argument('-f', '--file',help="specify an output file otherwise defaults to info.csv")
+parser.add_argument('--ac',help="specify area chair")
```
melisabok: I think this parameter is not being used

venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
```diff
@@ -0,0 +1,136 @@
...
+args = parser.parse_args()
+if args.file == None:
+    args.file = 'info.csv'
+
+## Initialize the client library with username and password
+client = Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+reviews_total =0
+reviews_complete = 0
+# paper_data[paper_num]['title'] = title
```
melisabok: we can remove commented code

venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
```diff
@@ -0,0 +1,136 @@
...
+    row = ['Paper Num', 'Title', 'Area Chair', 'reviews complete/total', 'Reviewer1', 'rating', 'confidence', 'Reviewer2', 'rating', 'confidence', 'Reviewer3', 'rating', 'confidence']
+    csvwriter.writerow(row)
+
+    for number in paper_data:
+        if args.ac == None or args.ac == paper_data[number]['AC']:
+            row = []
+            row.append(number)
+            row.append(paper_data[number]['title'].encode('UTF-8'))
+            row.append(paper_data[number]['AC'].encode('UTF-8'))
+            row.append("'{0}/{1}'".format(paper_data[number]['review_count'], paper_data[number]['reviewer_count']))
```
melisabok: instead of showing these two values in a single column. Could we split it in two? it is easier to sort in a excel:

- reviews_completed
- reviews_total

venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
```diff
@@ -0,0 +1,136 @@
...
+# paper_data[paper_num]['title'] = title
+# paper_data[paper_num]['reviewers']['rating'/'confidence'] = rating
+# paper_data[paper_num]['AC']= Area Chair id
+# paper_data[paper_num]['reviewer_count']= # of assigned reviewers
+# paper_data[paper_num]['review_count']= # of reviews complete
+paper_data = {}
+
+# fill in submission information
+print "Filling in submissions"
+submissions = client.get_notes(invitation=config.SUBMISSION)
```
melisabok: you need to get the blind submissions: config.BLIND_SUBMISSION

venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
pmandler: It is on line 119.  It allows you to output information of only one Area Chair.  Not necessary, but output seemed less overwhelming.

venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
pmandler: This is meant to be a description of the paper_data structure. Since it is multiple layers I like to have it all written out in one place. Maybe there is a better way to put it so it doesn't look like unused code.

venues/ICLR.cc/2018/Conference/python/programchairs/get-review-info.py
pmandler: Why does it need the blind submissions vs the actual submissions? I'm only getting the title so it doesn't matter that way. Is it because we could accidentally show authors and it's safer to deal with the blind submission?

melisabok: No, it is not that. 

I tested using live data and I got deleted papers. It seems that some paper were deleted only the blind submission but the original one still exists.

melisabok:
It looks good!! Thanks @pmandler 

@mspector  do you want to take a look?

PR #535, opened 2017-11-29
calculate confidence average, max and min

melisabok:


melisabok:
@mspector  when you have a change. The fix was deployed in live.

PR #536, opened 2017-11-30
Fix/author notes in tabs

mspector:
fixes a bug caused by passing the email address instead of user ID into the "term" parameter. If you pass in just an email address, the search query only finds papers with that email address in the authorids field.

### DIFF
--- a/venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
+++ b/venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
@@ -235,17 +235,12 @@ function renderContent(notes, submittedNotes, assignedNotePairs, userGroups, aut
   if (assignedPaperNumbers.length !== assignedNotes.length) {
     console.warn('WARNING: The number of assigned notes returned by API does not ' +
       'match the number of assigned note groups the user is a member of.');
-
-    console.log(assignedPaperNumbers);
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
```diff
@@ -89,11 +89,15 @@ function load() {
       );
     });
 
-    authorNotesP = Webfield.get('/notes/search', {
-      term: user.id,
-      group: CONFERENCE,
-      content: 'authors',
-      source: 'forum'
+    authorNotesP = Webfield.get('/user/profile', {
+      email: user.id
```
melisabok: what happens if the user.id is a tilde id?

mspector: is it possible for user.id to be a tilde id? I thought it was the login of the user, which should be an email address

melisabok: Ohh I see, it is the info of the logged user, you are right

melisabok:
I'm sorry, I shouldn't have merged this branch. I think you don't need to get the profile because you already have the tilde id in the UI, it is what we render un the menu. 

Try using user.profile.id as a term

PR #547, opened 2017-12-04
add decision tags

melisabok:
I run this in live, please don't merge it. I want to do the same for the other NIPS workshop. 

Let me know if you any bugs.

### DIFF
--- a/venues/NIPS.cc/2017/Workshop/Autodiff/python/create-decision-tags.py
+++ b/venues/NIPS.cc/2017/Workshop/Autodiff/python/create-decision-tags.py
@@ -1,0 +1,51 @@
+## Import statements
+import argparse
+from openreview import *
+import csv
+import sys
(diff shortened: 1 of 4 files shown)


melisabok:
@zbialecki: I applied the same changes to the MLTIS webfield.

melisabok:
Thanks! This is ready to merge.

pmandler:
For the MLITS version, why not use NIPS.cc/2017/Workshop/MLITS/-/Paper.*/Acceptance_Decision? 

There is no need to include: 
 +import csv
 +import sys
 +sys.path.insert(0, "./..")
 +import config

melisabok:
I'm sorry, I missed them, probably because they are not public. By the way those notes have the values: accept or reject. It is better to have in the tags options like: oral, poster or reject

melisabok:
is this ready to merge?

PR #552, opened 2017-12-11
naacl preprint infrastructure

mspector:


venues/naacl.org/NAACL/2018/Preprint/process/submissionProcess.js
```diff
@@ -0,0 +1,173 @@
...
+        var authorGroup = {
+          id: authorGroupId,
+          signatures: [CONF],
+          writers: [CONF],
+          members: note.content.authorids.concat(note.signatures),
+          readers: [CONF, authorGroupId],
+          signatories: [authorGroupId]
+        };
+
+        var withdrawPaperInvitation = {
```
melisabok: The withdraw process is the same as ICLR?

mspector: sorry, I missed this comment somehow.

the withdraw process is similar but not exactly the same as ICLR. When an author withdraws a paper here, both the blind note and the original are deleted (ddate is set). This is to prevent the user from having to manually delete both the original and the blind copy separately

venues/naacl.org/NAACL/2018/Preprint/python/admin-init.py
```diff
@@ -0,0 +1,50 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--username')
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+groups = {}
```
melisabok: I don't see reason to use dictionary here, using an array is fine. You only need to iterate through the array to post each group

venues/naacl.org/NAACL/2018/Preprint/python/admin-init.py
```diff
@@ -0,0 +1,50 @@
...
+
+groups[config.AUTHORS] = openreview.Group(config.AUTHORS, **config.group_params)
+
+groups[config.CONF] = client.get_group(config.CONF)
+groups[config.CONF].signatures = [client.signature]
+groups[config.CONF].add_webfield(config.WEBPATH)
+
+invitations = {}
```
melisabok: same for the invitations

venues/naacl.org/NAACL/2018/Preprint/python/admin-init.py
```diff
@@ -0,0 +1,50 @@
...
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+groups = {}
+
+groups[config.AUTHORS] = openreview.Group(config.AUTHORS, **config.group_params)
+
+groups[config.CONF] = client.get_group(config.CONF)
```
melisabok: here you can use the library helper to build the group.

mspector: originally, I wanted to keep the creation of the main groups separate from the setup of the conference. but I think you're right here, since I'm just going to be running this script anyway

mspector:
@melisabok, sorry I didn't answer your comments for so long. I have posted this on the live site, but if you have any more feedback I can still make changes

PR #556, opened 2017-12-23
Don't use global Promise object 

zbialecki:
Currently used in ICLR webfield. It causes the page not to load in IE11


melisabok:
Thanks @zbialecki !!!

PR #558, opened 2018-01-05
get assgigned papers from the anonreviewer groups

melisabok:
it is deployed in dev

### DIFF
--- a/venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
+++ b/venues/ICLR.cc/2018/Conference/webfield/conferenceWebfield_tabs.js
@@ -69,7 +69,7 @@ function load() {
     tauthor: true
   });
 
-  var assignedNotesP = Webfield.api.getSubmissions(WILDCARD_INVITATION, {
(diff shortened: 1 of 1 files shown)


melisabok:
Thanks!!!

zbialecki:
Everything seems to be working as expected. I'm a little confused by one thing though - why there are 2 papers listed under the My Tasks tab, but then My Assigned Papers has 5 papers (including the 2 from My Tasks)?

melisabok:
My Tasks shows the 2 papers that the review is missing. If you go to 'My Comments and Reviews' you can see 3 reviews done. So 2 missing reviews + 3 reviews done = 5 assigned papers.

melisabok:
My Tasks could be renamed to My Pending Tasks. That's confusing I think

zbialecki:
Ah I see, thanks! I think I was also confused because Red Paper had a comment, but not an official review. I agree that My Tasks could maybe be named better.

mspector:
One more question before we merge: am I supposed to be seeing the author IDs in the my assigned papers tab?

mspector:
I've tested this and deployed it to the live site. I'm on a spotty internet connection but when I can I'll send out the reviewer emails that Tara asked for

PR #563, opened 2018-01-09
update comment process to send notifications to ACs and reviewers

mspector:
Requested by the PCs. This has been tested pretty thoroughly, but notification emails always make me nervous. Please let me know if you have any suggestions


melisabok:
I think this is ok, it is difficult to see something wrong in the code, too much conditions. If you tested all the scenarios I would merge it

mspector:
@melisabok I did test it thoroughly, and I think it is ok. Could you merge it when you have a moment? then I'll run invitations.py from the app server to update the comment process functions.

PR #564, opened 2018-01-09
load profile information from NSF Grants

pmandler:


utils/load-NSF-profiles.py
```diff
@@ -0,0 +1,148 @@
...
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('-d','--directory', required=True, help = "the full path to the XML files")
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+
+# catch "not found" errors
+def get_profile(client, value):
```
melisabok: I moved this function to the openreview-py library, you can using it using the tools module:

https://github.com/iesl/openreview-py/blob/master/openreview/tools.py#L6

utils/load-NSF-profiles.py
```diff
@@ -0,0 +1,148 @@
...
+        profile = client.get_profile(value)
+    except openreview.OpenReviewException as e:
+        # throw an error if it is something other than "not found"
+        if e[0][0] != 'Profile not found':
+            print "OpenReviewException: {0}".format(e)
+            raise e
+    return profile
+
+
+def create_profile(client, email, first, last, tilde_id, verbose=False):
```
melisabok: there is a create profile in the library:

https://github.com/iesl/openreview-py/blob/master/openreview/tools.py#L17

It also checks if the profile already exists or not.

utils/load-NSF-profiles.py
```diff
@@ -0,0 +1,148 @@
...
+
+# for each new name/email, create a new profile
+def upload_profiles(client, investigators):
+    for email in investigators:
+        # try to find profile w/ email
+        profile = get_profile(client, email)
+        # if this email already has a profile, no need to do anything
+        # if not, if find profile w/~name
+        if profile == None:
+            tilde_id = "~{0}_{1}1".format(investigators[email]['first'], investigators[email]['last'])
```
melisabok: the create_profile that is in the tools module creates the tilde id so you don't need to do it here.

pmandler: That eliminates half the code!
Is there a way to add institutions to profiles?

melisa-bok-olx: Not now. I think we need to add a new revision to the profile with the new institution.  But this is not implemented yet.

pmandler:
Is this ready to run?

melisabok:
I think it is ok, I will merge it. Please, run it in the dev instance first.

PR #565, opened 2018-01-12
assign nonreaders instead of appending

mspector:
this is to make sure that the script can be run multiple times. it overwrites the nonreaders field of the official comments, but it's ok, because we never set this field to anything other than the authorids.


mspector:
@melisabok or @pmandler , do you mind taking a quick look at this, when you have a chance?

PR #567, opened 2018-01-19
scripts and modifications to existing scripts to update the reader op…

mspector:
…tions for public and official comments.

This PR addresses a problem with comment readership. When a commenter posts a comment with readership of "reviewers or higher", and the author of the paper is a reviewer of a different paper, then the author can actually see the comment.

We resolved this problem before by adding the author group as nonreaders to the comments where this applies, but that is not a permanent solution. With this PR, I propose we do the following:

1) we create paper-specific groups for Authors_and_Higher, Reviewers_and_Higher, and Area_Chairs_and_Higher.
2) we modify the comment invitations to accept those paper-specific values as dropdown options
3) we iterate through all comments and replace the readership with the appropriate paper-specific groups.


mspector:
@melisabok do we need to discuss this further, or can it be merged?

There was a bug (not introduced by these changes) that I fixed, but I had to fix it in this branch (because currently, the official comments are set up based on the process function in this branch).

PR #570, opened 2018-01-23
output csv file with all the info needed to make an acceptance decision

pmandler:
Script to create spreadsheet of all information need

### DIFF
--- a/venues/ICLR.cc/2018/Conference/python/programchairs/get-acceptance-status.py
+++ b/venues/ICLR.cc/2018/Conference/python/programchairs/get-acceptance-status.py
@@ -1,0 +1,126 @@
+#!/usr/bin/python
+
+"""
+Dump csv file of meta-review information to make final decision.
+"""
+import argparse
+import openreview
+import csv
+
+## Argument handling
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base url")
+parser.add_argument('--username')
+parser.add_argument('--password')
+parser.add_argument('--ofile', help="output file name - default to status.csv")
+args = parser.parse_args()
+
+def init_paper_info(client):
+    print "loading submissions"
+    submissions = client.get_notes(invitation='ICLR.cc/2018/Conference/-/Blind_Submission')
+    paper_info={}
+    for n in submissions:
+        paper_info[n.number] = {}
+        paper_info[n.number]['title'] = n.content['title']
+        paper_info[n.number]['AC_rec'] = ""
+        paper_info[n.number]['AC_conf'] = ""
+        paper_info[n.number]['AC_meta'] = ""
+        paper_info[n.number]['AC_name'] = ""
+        paper_info[n.number]['ave_score'] = 0
+        paper_info[n.number]['reviews'] = []
+        paper_info[n.number]['num_review'] = 0
+    return paper_info
+
+def get_activity(client, paper_info):
+    # load in all notes from conference
+    print "loading reviews"
+    limit = 1000
+    offset = 0
+    notes = []
+    batch = client.get_notes(invitation="ICLR.cc/2018/Conference/-/.*", limit=limit, offset=offset)
+    notes += batch
+    while len(batch) == limit:
+        offset += limit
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2018/Conference/python/programchairs/get-acceptance-status.py
```diff
@@ -0,0 +1,126 @@
...
+        paper_info[n.number]['num_review'] = 0
+    return paper_info
+
+def get_activity(client, paper_info):
+    # load in all notes from conference
+    print "loading reviews"
+    limit = 1000
+    offset = 0
+    notes = []
+    batch = client.get_notes(invitation="ICLR.cc/2018/Conference/-/.*", limit=limit, offset=offset)
```
melisabok: to avoid getting all the notes you can add the word 'Review' to the regex? Does it work?

melisabok:
Based on the google doc that Tara shared: https://docs.google.com/spreadsheets/d/1UorTnNv6G7aQGYJyKuor92IAII9DIS0LCtn8ZaXL36M/edit#gid=1478907816

The columns should be:


Paper ID | Paper title | AC recommendation | AC confidence | Average reviewer scores | Reviewer scores | Reviewer confidences | Assigned PC | DONE? | PC Decision | PC notes | AC Metareview | PC Metareview |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

melisabok:
Take a look at the reviews, they are displayed in a single column separated by semi colon.

pmandler:
invitation="ICLR.cc/2018/Conference/-/.*_Review” does not work. But invitation="ICLR.cc/2018/Conference/-/.*/Official_Review” does.
I could separate into two get_notes(), for the Official_Reviews and the Meta_Reviews.  It might be faster.

Pam 
___________________
Pam Mandler
Center for Data Science

melisabok:
I guess you need to get the submissions too. So you would need to change that way to get the data.

pmandler:
Ahh I didn’t notice the spreadsheet before.  What do you think Done means? Reviews done, MetaReviews Done, Final acceptance? We don’t have the PC Decisions an Notes, so those are blank columns - right?

Pam

melisabok:
I think you need to leave it blank and 'Assigned PC' as well. Please confirm the output with Tara.

melisabok:
Pam, did you try to run the script against the live site? it takes a lot of time to me and ends with an exception:

```
Operation timed out
```

I think we should load both type of review separately:

- https://openreview.net/notes?invitation=ICLR.cc/2018/Conference/-/Paper.*/Meta_Review (you don't to paginate, it is the same size as the submissions)
- https://openreview.net/notes?invitation=ICLR.cc/2018/Conference/-/Paper.*/Official_Review (get all the pages)

and we you do 

```
client.get_profile
```

save the result in dictionary to store the profiles, next time that you need to get the profile check in the dictionary first. This will reduce a lot of network calls.

pmandler:
I've made the suggested changes and tried it out on the live site without any exceptions.

melisabok:
great! It works for me too, one more thing: when I open the file in the excel the metareview text is not displayed correctly. I think because it is a multiline text and we should wrap it with double quotes. Do you remember how did you do last year?

pmandler:
That's odd, because it works for me.  We didn't do anything special with it last year.  Shall I add the double quotes and you can see if it makes a difference?

pmandler:
Let me know when I can send this to Tara

melisabok:
You were right, it is an excel issue. 
Google docs works ok: https://docs.google.com/spreadsheets/d/1TZ4WLyYYs5uQvFgyrjVnXhZskkDf4KLR2f2gt1c7BcM/edit#gid=1296491265

pmandler:
Turns out if you resize the cell vertically, the text is there.

melisabok:
Ohhhh, thanks Pam!

PR #572, opened 2018-01-25
Dev draft of MIDL conference

pmandler:


pmandler:
This conference has two tracks, the conference track and the abstract track. The only difference at this point is the expected length of pdf submissions which is described in the submission form.

PR #578, opened 2018-01-29
output xml of information for accepted submissions

pmandler:


pmandler:
Oddly, just adding the check for the Value Errors seems to prevent them.

melisabok:
We are not sure if after 3 times the request is going to work. May be we are doing too many requests. You could reduce the amount of calls getting all the author groups and profile at once:

Author groups:

```
https://openreview.net/groups?id=ICLR.cc/2018/Conference/Paper.*/Authors
```

Profiles:

```
POST /user/profiles
{
emails: [list of all the author emails]
}
```
The last one, I think is not implemented in the library, you could add it.

melisabok:
It looks good!

PR #584, opened 2018-02-01
Feature/researcher profiles

mspector:
@pmandler , could we go over this sometime? I can talk you through everything


pmandler:
This got lost in my inbox.  Want to get together early next week?

mspector:
Thanks Pam. Honestly it might not be worth the time... since these are all
new files, and they're very likely to change in the future, I might just
merge these in myself. On the other hand, if you're particularly curious
about how the ECCV demo works, I would be happy to go over it with you. If
not, then I don't think the review is worth the time.

pmandler:
I'll skip the review then.  Curiosity vs time.

PR #595, opened 2018-02-12
add recruit reviewers script

mspector:


mspector:
@pmandler this has been used already, so a review would be kind of unnecessary. i'm going to go ahead and merge it so it's part of the history

PR #596, opened 2018-02-12
significantly increase the size limit of tldr field

mspector:


melisabok:
Are you sure that this fields is still named: 'Too long didn't read: : a short sentence describing your paper?'

mspector:
I'm not sure I understand your question. The title and description have not changed.

I'm increasing the size because, for some reason, many of the original submissions to the Conference track exceed the size limit. There may be something wrong with the validation, but I haven't had the time to check it. The old size limit was causing transfers from the conference to the workshop to fail, so I changed the size limit.

melisabok:
It is supposed to be a field with a short description, increasing the length to 5000 removes this characteristic.

mspector:
ok, i'll make it smaller.

melisabok:
Probably the issue comes from the Conference submission invitation has a different regex:

"value-regex": "[^\n]{0,500}"

https://openreview.net/invitations?id=ICLR.cc/2018/Conference/-/Submission

mspector:
@melisabok is this ok too? the sizes now match each other

PR #597, opened 2018-02-13
Feature/bid webfield enhancements

zbialecki:
Add sorting option to Add Bid webfield that allows the user to sort by affinity score. Also label conflicts with a red tag. 

@mspector I created a new js file to make testing easier (so I could js run the update-invitation script), but I also transferred the new code over to the template file. I haven’t tested out generating the webfield from the template file though.


mspector:
@zbialecki looks good to me. I have a couple comments:

1) could we set the default sorting option to "affinity score"?
2) it would be nice if the user could click on the subject area filter field to display all the options. right now it seems like you have to start typing something to get anything to show up.

zbialecki:
Yeah, no problem. I updated the both PRs to change affinity score to the default sort and fix the bug with the dropdown. Let me know if there's anything else.

mspector:
thanks @zbialecki !

PR #599, opened 2018-02-13
ICLR Workshop - post early acceptance decisions

pmandler:
config and invitations - adds Acceptance_Decision invitation.  Unfortunately it allows multiple responses - how do I change it so it only accepts one per paper?
post-acceptance-decisions.py based on the script from Conference.  Added ability to output decisions from Conference to help with early decisions.

### DIFF
--- a/venues/ICLR.cc/2018/Workshop/python/invitations.py
+++ b/venues/ICLR.cc/2018/Workshop/python/invitations.py
@@ -69,7 +69,7 @@ invitation_configurations = {
         'byPaper': False,
         'byForum': False,
         'params': config.acceptance_decision_params,
-        'invitees': []
+        'invitees': [config.PROGRAM_CHAIRS]
     },
     'Add_Bid': {
         'tags': True,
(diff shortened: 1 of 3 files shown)


mspector:
@pmandler the way we make an invitation accept only one paper is with the process function. We basically tell the process function to add the user to the list of noninvitees. But that may not be necessary in this case - this invitation is getting filled out either over the API with a script, or from the GUI by a program chair, who will know better than to post multiple decisions.

venues/ICLR.cc/2018/Workshop/python/invitations.py
```diff
@@ -65,6 +65,12 @@
         'signatures': [config.PROGRAM_CHAIRS],
         'params': config.meta_review_params
     },
+    'Acceptance_Decision': {
+        'byPaper': False,
+        'byForum': False,
+        'params': config.acceptance_decision_params,
+        'invitees': []
```
mspector: you could add the program chairs group as "invitees" here

PR #601, opened 2018-02-14
add get pdfs script

mspector:


pmandler:
The check if the directory exists should happen once before the for loop.

PR #624, opened 2018-03-06
Fix / UAI reviewer bid TPMS

zbialecki:
Fixes some more bugs with UAI review bid interface and adds new features like sorting submissions by TPMS scores.

Not quite ready to merge, but I wanted to open a PR for review and testing.


zbialecki:
Currently all changes are saved in venues/auai.org/UAI/2018/webfield/bidWebfield_updated.js, however once everything is done I'll move them over to the webfield template file.

mspector:
Thanks a lot Zach! this is looking good to me. Everything seems to be working propery (except the bug with the search box that Melisa pointed out, but I think you knew about that).

By "the webfield template file", you just mean the other webfield "bidWebfield.js", right? (as opposed to, for example, some handlebars template that lives in the system itself). Or did you mean that?

mspector:
@zbialecki , after my meeting with the UAI program chairs, here's some feedback for the bidding interface:

1. I think there's a bug in the subject area filter. Here are the symptoms: when they enter a certain subject area, the interface only finds papers that have that subject area exclusively - in other words, if a paper has multiple subject areas, the filter seems to ignore them.

2. the UAI PCs have asked if it would be possible to make the scores visible in the bidding interface. They say that this could be useful because it would let people see the magnitude of the difference between the scores. I don't think this is absolutely critical, but if it's an easy change, I think they would like it

3. In terms of the sorting by score: they asked that the "Default" option be alphabetical, so I just committed a small change to make that work. Let me know if there's a better way to do this.

mspector:
@zbialecki I'm going to merge this for now; we can keep tweaking in another branch. is that ok?

venues/auai.org/UAI/2018/webfield/bidWebfield_updated.js
```diff
@@ -391,7 +391,7 @@ function renderContent(validNotes, tagInvitations, metadataNotesMap) {
             }
             Webfield.ui.searchResults(sortedNotes, submissionListOptions);
           } else {
-            Webfield.ui.searchResults(notes, submissionListOptions);
+            Webfield.ui.searchResults(_.sortBy(notes, n => n.content.title.toLowerCase()), submissionListOptions);
```
zbialecki: This will cause problems in for any users of IE. Arrow functions should only be used on the server side.

PR #625, opened 2018-03-06
Uai2018/matching updates

mspector:
@melisabok this PR contains the README instructions for setting up the matching system in UAI 2018


melisabok:
Thanks! I will try to run it right and give you feedback.

PR #627, opened 2018-03-08
add a loop to get all papers

mspector:
@melisabok is this what you had in mind? let me know


mspector:
@melisabok ok, I updated the tools module here:

https://github.com/iesl/openreview-py/pull/84

melisabok:
You need to do the same in match-reviewers.py. Line 65 and 66, and I guess line 91 too.

PR #631, opened 2018-03-09
add get-pdfs script

mspector:


venues/auai.org/UAI/2018/python/get-pdfs.py
```diff
@@ -0,0 +1,29 @@
...
+print 'connecting to {0}'.format(client.baseurl)
+
+submission_invitation = 'auai.org/UAI/2018/-/Blind_Submission'
+submissions = client.get_notes(invitation=submission_invitation)
+
+if not os.path.exists('../pdfs'):
+    os.makedirs('../pdfs')
+
+for n in submissions:
+    pdf_url = 'https://openreview.net{0}'.format(n.content['pdf'])
```
pmandler: Shouldn't this be client.baseurl instead of https://openreview.net?

mspector: yes, thanks pam!

PR #633, opened 2018-03-13
Change UAI bid webfield to use local search instead of API

zbialecki:
This PR changes the reviewer bid webfield to use a client side search instead of the /notes/search API. This should improve the search results as well as the speed of the search. It also fixes the issue the search API had with subject areas.

Requires https://github.com/iesl/openreview/pull/897


melisabok:
@mspector  you merged it but this is not ready, it depends on another PR

PR #648, opened 2018-03-18
ignore hard constraints while computing final score

mspector:


melisabok:
Thanks @mspector ! Thanks for fixing the typos 👍

mspector:
@melisabok sorry for the delay, the fix has been applied to the reviewers script now

melisabok:
Ok, I will merge it. But we must to read the scores from the matching result

melisabok:
and we need to replicate the same changes in ECCV. I don't agree doing this.

mspector:
@melisabok are you worried that we're replicating too much code? I can generalize this so that we use the same script for all four cases (UAI ACs and Reviewers, ECCV ACs and Reviewers)

mspector:
FYI to anyone else following this: @melisabok and I discussed this, and I'm going to generalize these scripts to avoid repetition. I'm keeping open this PR for now, so that we can reference it. I will close it when I have a new PR ready.

PR #649, opened 2018-03-18
Feature/setup eccv matching

melisabok:
DON'T MERGE.

Initial commit for the setup of ECCV.


venues/cv-foundation.org/ECCV/2018/Conference/python/upload-areachairs.py
```diff
@@ -0,0 +1,38 @@
...
+    reader = csv.reader(f)
+    reader.next()
+    members = []
+    for line in reader:
+        first = line[0].strip()
+        middle = line[1].strip()
+        last = line[2].strip()
+        email = line[3].lower().strip()
+        try:
+            profile = openreview.tools.create_profile(client, first = first, middle = middle, last = last, email = email)
```
melisabok: it shouldn't create any profile, because all the profiles should come from researcher. So, I think we should change the call from create_profile to get_profile by email

PR #650, opened 2018-03-19
Fix/eccv matching

melisabok:


venues/cv-foundation.org/ECCV/2018/Conference/python/submit-papers.py
```diff
@@ -38,6 +38,9 @@ def create_cv_note(title, abstract, authors, authorids, subject_areas, paper_id)
         authorids = [l.strip() for l in line[6].split(';')]
         subject_areas = [l.strip().replace('*','') for l in line[8].split(';')]
 
-        note = create_cv_note(title, abstract, authors, authorids, subject_areas, number)
-        posted_n = client.post_note(note)
-        print posted_n.content['title']
+        if number not in ['497', '592', '2388', '3200']:
```
mspector: could you add a comment to this so that we remember why these numbers are excluded?

melisabok: Done!

PR #652, opened 2018-03-19
Fix/iclr get accepted

pmandler:
lines 127-128 are the main change.


melisabok:
We have conflicts here.

PR #662, opened 2018-03-22
update script for area chair recommendations

mspector:
updates the UAI metadata to include area chair recommendation scores


venues/auai.org/UAI/2018/python/update-recommendations-metadata.py
```diff
@@ -0,0 +1,57 @@
...
+    this is a linear mapping between the recommended reviewer's position in the recommendation
+    list (its "rank") and a "recommendation score" for the reviewer and that paper.
+    '''
+
+    for rank, entry in enumerate(recommendations):
+        userid = entry.tag
+        for entry in user_scores:
+            if entry['userId'] == userid:
+                print entry['userId'], userid
+                entry['scores']['recommendation-score'] = float(rank_score_map[rank])
```
melisabok: You have different keys:

recommendation-score
recommendation_score

melisabok: It would be nice to call the score using Json style:

recommendationScore

venues/auai.org/UAI/2018/python/update-recommendations-metadata.py
```diff
@@ -0,0 +1,57 @@
+import openreview
```
melisabok: Why don't we have a setup-metadata.py script only?

It creates or update metadata notes using all the inputs:

-tpms scores files
-bids
-recommendations
....

melisabok: instead of having one script per type of score

venues/auai.org/UAI/2018/python/update-recommendations-metadata.py
mspector: Ah, good catch! thanks!

venues/auai.org/UAI/2018/python/update-recommendations-metadata.py
mspector: ok, good idea

mspector: i think this is good, but i haven't had time to implement a single updating script yet. I'm going to close this in favor of this branch (https://github.com/iesl/openreview-scripts/pull/664) which contains the changes here.

PR #672, opened 2018-03-30
IEEE scripts for creating paper groups, review invitations etc

pmandler:
init-submissions.py - add comment invitation, PC group and  reviewer group
get-submissions.py - output submission info
create-invitations.py - for each paper create paper group, authors, reviewers etc and create review invitation
assign-reviewers.py - assigns reviewer to paper after doing basic checks for conflicts.


venues/IEEE.org/2018/ITSC/python/assign-reviewers.py
```diff
@@ -0,0 +1,172 @@
...
+        nonreaders=conflict_list,
+        signatories=[new_reviewer_id])
+    client.post_group(new_reviewer)
+    print "assigned user %s to group %s" % (reviewer_email, new_reviewer_id)
+    return new_reviewer
+
+# does some parameter checking, checks for conflicts
+# check if reviewer already assigned to paper
+# if not then create Anonymous group for the reviewer and add to reviewer group
+def assign_reviewer(reviewer_email, paper_number):
```
melisabok: There is an assign function in the library, could try to use it here?

https://github.com/iesl/openreview-py/blob/master/openreview/tools.py#L275

So we don't need to copy/paste the same code in all the conferences.

venues/IEEE.org/2018/ITSC/python/create-invitations.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/python
```
melisabok: this seems to be a create-review-invitations script?

pmandler: Yes, I'll rename it.

venues/IEEE.org/2018/ITSC/python/assign-reviewers.py
```diff
@@ -0,0 +1,122 @@
...
+        conflict_list += note.content['authorids']
+
+    if reviewer_conflicts(reviewer_email, paper_number, conflict_list):
+        return True
+
+    reviewer_id = tools.get_profile(client, reviewer_email)
+    if reviewer_id is None:
+        reviewer_id = reviewer_email
+    else:
+        reviewer_id = reviewer_id.id
```
mspector: I made a PR on openreview-py earlier today that does this inside the tools.assign function. I'll leave this PR open for right now. If you have a chance to make the change then great, but this should work either way.

PR #678, opened 2018-04-05
Feature/new matching api

mspector:
How to match ECCV reviewers:

1. Ensure that all researcher profiles that we know about have equivalent profiles in OpenReview.
2. Run upload-reviewers.py. This will:
- Detect any missing profiles (there should be none, when we run this in the real conference)
- Create the cv-foundation.org/ECCV/2018/Conference/Reviewers group
3. Run setup-matching.py. This will create the new, reviewer-specific metadata invitation.
4. Run download-profiles.py to get the profiles.pkl file
5. Run download-reviewer_profiles.py to get the cmt_conflicts.pkl file
5. Run setup-metadata-reviewers.py. This will generate the metadata required to do the match.
6. Run match-reviewers.py. Make sure to adjust the configuration directly in the file, if necessary

### DIFF
--- a/venues/cv-foundation.org/ECCV/2018/Conference/python/setup-matching.py
+++ b/venues/cv-foundation.org/ECCV/2018/Conference/python/setup-matching.py
@@ -36,7 +36,7 @@ metadata_inv = client.post_invitation(openreview.Invitation(**{
 }))
 
 metadata_reviewers_inv = client.post_invitation(openreview.Invitation(**{
-    'id': 'cv-foundation.org/ECCV/2018/Conference/-/Paper_Metadata/Reviewers',
+    'id': 'cv-foundation.org/ECCV/2018/Conference/Reviewers/-/Paper_Metadata',
     'readers': [
         'cv-foundation.org/ECCV/2018/Conference',
         'cv-foundation.org/ECCV/2018/Conference/Program_Chairs'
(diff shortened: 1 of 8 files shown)


venues/cv-foundation.org/ECCV/2018/Conference/python/setup-matching.py
```diff
@@ -35,6 +35,28 @@
     }
 }))
 
+metadata_reviewers_inv = client.post_invitation(openreview.Invitation(**{
+    'id': 'cv-foundation.org/ECCV/2018/Conference/-/Paper_Metadata/Reviewers',
```
melisabok: Why not call the group like this:

cv-foundation.org/ECCV/2018/Conference/Reviewers/-/Paper_Metadata?

mspector: good idea. that would work

melisabok:
I found a conflict that is a gmail address?

screen shot 2018-04-05 at 7 07 51 pm

melisabok:
Well the reviewer is the author of the paper :)

melisabok:
I could run the matcher changing the maxpapers to 20. I see that finalScore should be the same than tpmsScore, see image attached:

screen shot 2018-04-05 at 8 24 42 pm

It seems that finalScore is the tpmsScore divided by 3.

melisabok:
After ECCV we should discuss the scores and constraints in more detail. I’m not yet sure if they should be separated or not. Thanks!

mspector:
I'm working on a branch where the conflicts and scores are separated. it will be separate from this branch, so this can continue working as intended

melisabok:
Ok, but that change implies to change the matching code, right?

mspector:
Yes, I have a corresponding branch in the matching repo too.

PR #687, opened 2018-04-14
MIDL Conference - add field to review

pmandler:
Need to rerun create-review-invitations.py to propagate change.

### DIFF
--- a/venues/MIDL.amsterdam/2018/Conference/python/config.py
+++ b/venues/MIDL.amsterdam/2018/Conference/python/config.py
@@ -213,8 +213,18 @@ review_content = {
         'description': 'Please provide an evaluation of the quality, clarity, originality and significance of this work, including a list of its pros and cons (up to 5000 chars).',
         'required': True
     },
-    'rating': {
+    'special_issue': {
         'order': 3,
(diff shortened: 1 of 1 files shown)


venues/MIDL.amsterdam/2018/Conference/python/config.py
```diff
@@ -213,8 +213,18 @@
         'description': 'Please provide an evaluation of the quality, clarity, originality and significance of this work, including a list of its pros and cons (up to 5000 chars).',
         'required': True
     },
-    'rating': {
+    'special_issue': {
         'order': 3,
+        'description': 'Is this paper suited for the Medical Image Analysis Special Issue?',
+        'value-radio': [
+            'Definitely',
```
melisabok: What is the difference between definitely and yes?

pmandler: No idea, but that is what was requested.
"The option should be ‘Is this paper suited for the Medical Image Analysis Special Issue?’ with three options Definitely, Yes, No. "

melisabok: I’m sorry I missed that part of the email.

PR #693, opened 2018-04-24
fix bug in recruitment process function

mspector:
Originally I was trying to (1) get the paper by the note's forum, and then (2) use the paper.number to determine the Paper#/Reviewers/Accepted group to add the new user to.

later, I changed the reviewer_invitation notes such that they do not respond to any particular forum. I did this to avoid having the button show up in the UI.

This caused the call in (1) above to retrieve the reviewer_invitation note itself, causing paper.number to be wrong.

### DIFF
--- a/venues/MIDL.amsterdam/2018/Conference/process/recruitReviewerProcess.js
+++ b/venues/MIDL.amsterdam/2018/Conference/process/recruitReviewerProcess.js
@@ -1,64 +1,63 @@
 function() {
-
   console.log('process initiated');
+
+  var invitationSplits = note.invitation.split('/');
+  var paperNumber = invitationSplits[4];
+  console.log('paperNumber identified: ' + paperNumber);
+
   var or3client = lib.or3client;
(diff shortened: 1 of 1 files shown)


venues/MIDL.amsterdam/2018/Conference/process/recruitReviewerProcess.js
```diff
@@ -1,64 +1,63 @@
...
+      };
+      var messageP = or3client.or3request(or3client.mailUrl, orteamMessageBody, 'POST', token);
+
     }
-  })
-  .then(result=>done())
-  .catch(error=>done(error))
 
-  return true
+    Promise.all([acceptP, declineP, messageP])
```
melisabok: promises variables are defined inside the if that is not the scope of the Promise.all. I'm not sure if this is working.

mspector: my tests were working (thought it's possible I missed something). also, I think this will be ok, because the promise variables are instantiated in both branches of the if statement. either way, the promises will be made

PR #694, opened 2018-04-25
Feature/second round ac match

mspector:
requires https://github.com/iesl/openreview-matching/pull/22

Summary of changes:
- generalized "download-profiles.py" script to handle both reviewers and area chairs
- added "download-score-maps-2nd-areachair.py" script to process the "flattened" TPMS scores
- added more parameters to the "match-areachairs.py" script to allow appending one assignment to another
- added AC-specific metadata invitation to "setup-matching.py" script
- added "setup-metadata-acs2.py" script for setting up AC-specific metadata


mspector:
@melisabok not urgent, but could you check this out when you have a chance? this is what I used to implement the 2nd round of AC matches

PR #698, opened 2018-04-27
Fix program chair console and add it to the home page

melisabok:


mspector:
is this deployed already?

melisabok:
Yes! I'm not sure if the pcs are aware of this console so I added to the Task tab.

mspector:
Thanks! They should be aware of it, but I didn't realize that it wasn't in the tasks tab! Thanks Melisa

PR #702, opened 2018-05-04
Fix/update comment process

mspector:
Some Official Comments have been made that have only the Reviewers group set as readers. the PCs would like to force these comments to be readable by the ACs as well.

To solve this, we should (1) add the AC groups to the reviewer groups, (2) update the Official_Review invitations so that the AC groups are noninvitees, and (3) update the Official_Comment process function so that the ACs don't get double emails.

To do (1), run the following:

```
papers = client.get_notes(invitation='auai.org/UAI/2018/-/Blind_Submission')
for paper in papers:
    client.add_members_to_group(client.get_group('auai.org/UAI/2018/Paper{}/Reviewers'.format(paper.number)),
                                'auai.org/UAI/2018/Paper{}/Area_Chairs'.format(paper.number))

```

To do (2) and (3), run `python setup-invitations.py Official_Review Official_Comment`. It's probably worth doing this from one of the app instances, for the sake of speed.

### DIFF
--- a/venues/auai.org/UAI/2018/process/commentProcess.js
+++ b/venues/auai.org/UAI/2018/process/commentProcess.js
@@ -62,7 +62,7 @@ function(){
         if (checkReadersMatch(/auai.org\/UAI\/2018\/Paper[0-9]+\/Reviewers/)){
           promises.push(or3client.or3request(or3client.mailUrl, reviewer_mail, 'POST', token));
         }
-        if (checkReadersMatch(/auai.org\/UAI\/2018\/Paper[0-9]+\/Area_Chairs/)){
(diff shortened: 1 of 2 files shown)


pmandler:
Adding the ACs to the Reviewers group blurs the meanings of the groups.  Why not change the commentProcess to add ACs if checkReadersMatch(...Reviewers)?

mspector:
The reason we're adding the AC to the reviewers group isn't to make sure that they get emails, it's to make sure that they can actually view the note. 

It's unclear what you mean by "add ACs" in this context, but if you mean "add them to the email recipients", this wouldn't solve the main problem. If you mean "add them to the reviewers group" (as part of the process function), then this would work for future comments, but we want to retroactively apply this to older comments too.

Makes sense? Let me know if you have other questions

PR #707, opened 2018-05-07
MIDL allow authors to see completed reviews

pmandler:


mspector:
@pmandler @melisabok what's happening with this PR? should it be merged?

pmandler:
It's not going to be used as a chron job as originally thought, but I did use it a couple times so please go ahead and merge it.

PR #710, opened 2018-05-08
Feature/nampi recruitment

mspector:


venues/ICML.cc/2018/Workshop/NAMPI/process/recruitReviewerProcess.js
```diff
@@ -0,0 +1,36 @@
...
+  var REVIEWERS_DECLINE_ID = REVIEWERS_ID + '/Declined'
+  var HASH_SEED = "2810398440804348173"
+
+  var hashKey = or3client.createHash(note.content.email, HASH_SEED);
+
+  if(hashKey == note.content.key) {
+    if (note.content.response == 'Yes') {
+      console.log("Invitation replied Yes")
+      //if a user is in the declined group, remove them from that group and add them to the reviewers group
+      or3client.removeGroupMember(REVIEWERS_DECLINE_ID, note.content.email, token);
```
melisabok: promises are not linked

venues/ICML.cc/2018/Workshop/NAMPI/process/recruitReviewerProcess.js
```diff
@@ -0,0 +1,36 @@
...
+      or3client.addGroupMember(REVIEWERS_ID, note.content.email, token)
+      .then(result => done())
+      .catch(error => done(error));
+    } else if (note.content.response == 'No'){
+      console.log("Invitation replied No")
+      //if a user is in the reviewers group, remove them from that group and add them to the reviewers-declined group
+      or3client.removeGroupMember(REVIEWERS_ID, note.content.email, token)
```
melisabok: same here

venues/ICML.cc/2018/Workshop/NAMPI/python/config.py
```diff
@@ -0,0 +1,101 @@
...
+PROGRAM_CHAIRS_ID = CONFERENCE_ID + '/Program_Chairs'
+REVIEWERS_ID = CONFERENCE_ID + '/Reviewers'
+REVIEWERS_INVITED_ID = REVIEWERS_ID + '/Invited'
+REVIEWERS_DECLINED_ID = REVIEWERS_ID + '/Declined'
+
+
+'''
+Group definitions.
+'''
+CONFERENCE = openreview.Group(**{
```
melisabok: we don't need uppercase variable names

venues/ICML.cc/2018/Workshop/NAMPI/python/config.py
```diff
@@ -0,0 +1,101 @@
...
+2018-05-08: setup workshop recruiting infrastructure
+
+
+'''
+
+import openreview
+from openreview import invitations
+import os
+
+CONFERENCE_ID = 'ICML.cc/2018/Workshop/NAMPI'
```
melisabok: when do you create each group of the path?

melisabok: if it is already created we should check that exists first.

venues/ICML.cc/2018/Workshop/NAMPI/python/setup.ipynb
```diff
@@ -0,0 +1,90 @@
+{
```
melisabok: please don't push a notebook, I can not follow the code here

melisabok: if it is 'one-time-use' then we don't need to commit the code then. 

I can not do code review of a python notebook.

pmandler:
Is there another way of sharing the notebooks? I would find it useful to see how another conference handled certain things.

venues/ICML.cc/2018/Workshop/NAMPI/python/config.py
mspector: I don't have strong opinions about this, but there is at least some argument for having uppercase variable names in this context:

https://www.python.org/dev/peps/pep-0008/#constants

it talks about using uppercase variable names "at the module level". It's not a perfect comparison, but I do tend to think of the /python directories as being a sort of module for the conference.

If you disagree I don't mind changing it. just wanted to share my reasons

melisabok: Well I see as constants, variables that won't change like the strings to define the group ids or invitations ids. 

These are instances of an object that will change in the future. And I don't think the conference group is a constant. May be we can define here the base config of the conference group but not the conference group itself, because this group will change in the future or it is already created.

mspector:
Good point - I'll change this to reflect the group parameters instead of
the object itself.

venues/ICML.cc/2018/Workshop/NAMPI/python/config.py
melisabok: I'm not sure if we need to save the group config as a constant. May be we just need to create the group in the setup script and that's all.

mspector: Good point - I'll change this to reflect the group parameters instead of
the object itself.

mspector: oops, our responses got out of sync.

Let me think about this a little more. maybe you're right

mspector: Now that I've gathered my thoughts:

The reason I did it this way is to avoid creating a setup script altogether. Setup scripts are run once and never run again, and they rarely stay up to date. They're also unwieldy: a lot of the time we want to make small changes in the setup but can't run the setup script because it would break other parts of the conference.

by putting object definitions (groups, invitations, etc) in the config file, users can import the file and then manipulate the contents as they see fit. Yes, there is a risk that the objects in the config file won't be up-to-date, but there is the same risk with a setup script.

in my opinion, the solution to this is a cultural one, not a technical one: we have to make sure that admins / workflow engineers understand that the config file represents the state that the conference is supposed to be in, and that if they make changes, they need to commit them to the repo. Of course they could make changes outside of this script, but at least this way there is a single location where the parameters are *supposed* to be stored.

melisabok: We can talk about this afternoon if you want and discuss the new changes that you want to do.

mspector: Ok, that sounds good. I'll send you an email to figure out a good time.

mspector:
@melisabok ok, the object variable names have been lowercased.

do you have a suggestion for a better name for the config.py file?

maybe objects.py? or even nampi.py?

venues/ICML.cc/2018/Workshop/NAMPI/python/functions.py
```diff
@@ -0,0 +1,39 @@
...
+        name = first,
+        accept_url = url + "Yes",
+        decline_url = url + "No"
+    )
+
+    # send the email through openreview
+    response = client.send_mail(config.RECRUIT_MESSAGE_SUBJ, [email], personalized_message)
+
+    if 'groups' in response and response['groups']:
+        reviewers_invited = client.get_group(config.REVIEWERS_INVITED_ID)
```
melisabok: This is confusing, why you didn't use:
```
 client.add_members_to_group(config.reviewers_invited, response['groups'])
```

melisabok: or:
```
 config.reviewers_invited = client.add_members_to_group(config.reviewers_invited, response['groups'])
```

venues/ICML.cc/2018/Workshop/NAMPI/python/setup.py
```diff
@@ -0,0 +1,46 @@
...
+
+# In[3]:
+
+
+'''
+Post the reviewer groups. Warning! This will reset all members!
+
+TODO: make a function to post a group without overwriting members.
+'''
+reviewers = client.post_group(config.reviewers)
```
melisabok: Are you assuming here that config.conference was already created?

melisabok:
I like nampi.py!

PR #716, opened 2018-05-20
Fix bug with homepage webfield template

zbialecki:
Tabs that were initially hidden were not showing properly when users submitted a paper.

The only active conference that has been fixed is NAMPI 2018 Workshop. The others will have to be patched individually.


zbialecki:
I checked all the other venues on live that are currently accepting submissions, and all use the basic template so they are unaffected by this issue.

PR #721, opened 2018-05-31
Fix/Replace tauthor with content.authorids

zbialecki:
Switch from getting authored notes with /notes/search to using the new content.authorids param

Depends on https://github.com/iesl/openreview/pull/985


melisabok:
Thanks for this change, but we don’t need to change past conferences. Changing the general webfield code is ok

zbialecki:
If it's OK with you, I'd like to leave those changes to the old webfields in there. Those webfields aren't exactly the same as the tabs template and in the future if they are copied for a new conference I want to make sure this change is in there.

PR #728, opened 2018-06-26
Feature/mag import

mspector:


venues/Microsoft.com/Academic_Graph/process/magTransform.js
```diff
@@ -0,0 +1,78 @@
+function (note) {
+  var _ = require('lodash');
```
melisabok: does it work this function?

I'm not sure if the require command would work, BTW we are trying to avoid the use of lodash and try to solve all the operation using plain javascript:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects

mspector: ah ok. I believe it does work, but I'll change it to plain js

mspector:
closing this for now. @melisabok we should discuss again when we talk about doing the mass data upload

PR #733, opened 2018-07-13
use new update_profile() and update logic

pmandler:


pmandler:
I'd still like to add coauthor information where appropriate.

PR #734, opened 2018-07-23
use tag details

melisabok:
Depends on https://github.com/iesl/openreview/pull/1042

No need to fix old Bid webfields. We can test this when we configure ICLR 2019.


melisabok:
This is ready to merge too

PR #736, opened 2018-08-01
add ICLR 19 scripts

mspector:
@melisabok should I avoid committing these webfields?

### DIFF
--- a/venues/ICLR.cc/2019/process/submissionProcess.js
+++ b/venues/ICLR.cc/2019/process/submissionProcess.js
@@ -1,0 +1,87 @@
+function() {
+  var or3client = lib.or3client;
+  console.log('submission process');
+
+  var SHORT_PHRASE = 'ICLR 2019';
+  var CONFERENCE_ID = 'ICLR.cc/2019/Conference';
+  var PROGRAM_CHAIRS_ID = CONFERENCE_ID + '/Program_Chairs';
+  var REVIEWERS_ID = CONFERENCE_ID + '/Reviewers';
+  var AREA_CHAIRS_ID = CONFERENCE_ID + '/Area_Chairs';
+  var BLIND_SUBMISSION_ID = CONFERENCE_ID + '/-/Blind_Submission';
+
+  var PAPER_GROUP_ID = CONFERENCE_ID + '/Paper' + note.number;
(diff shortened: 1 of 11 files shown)


melisabok:
I don't know. 

I think they are always out of sync. 

but it may happen that we can do a mess with the object in the database.

venues/ICLR.cc/2019/process/submissionProcess.js
```diff
@@ -0,0 +1,87 @@
+function() {
+  var or3client = lib.or3client;
+  console.log('submission process');
```
melisabok: Could you prepare a submission process with only email notifications?

mspector: I think @mohituniyal has that file already. he could probably just commit it to this branch

mohituniyal: I have committed the file. @mspector and @melisabok  Please take a look as I do not have a 100% confidence in the change.

melisabok: I would keep both versions, one that creates the blinded notes for testing purposes and the other one that only send emails, no revision invitation, for the live site.

venues/ICLR.cc/2019/process/submissionProcess.js
```diff
@@ -0,0 +1,46 @@
...
+
+  var SHORT_PHRASE = 'ICLR 2019';
```
melisabok: Remove the var that we don't use

venues/ICLR.cc/2019/process/submissionProcess.js
```diff
@@ -0,0 +1,46 @@
...
+    reply: {
+      referent: note.id,
+      forum: note.forum,
+      content: invitation.reply.content,
+      signatures: invitation.reply.signatures,
+      writers: invitation.reply.writers,
+      readers: invitation.reply.readers
+    }
+  };
+  or3client.or3request(or3client.inviteUrl, revisionInvitation, 'POST', token)
```
melisabok: We can set all the authorids as writers of the note, so there is no need to create an add revision invitation

PR #739, opened 2018-08-10
Turn on localSearch option to reduce extra API calls

zbialecki:


mspector:
this is ok with me, but I changed it to API search mode for two reasons: 1) the local search doesn't seem to pick up on subject areas in the text field (which is nice, but ultimately not critical because of the subject area filter), and 2) I'm under the impression that local search won't search beyond the batch of papers curently loaded on the homepage (since we use pagination).

@melisabok what do you think?

venues/ICLR.cc/2019/webfield/homepage.js
```diff
@@ -333,7 +333,7 @@ function renderContent(notes, userGroups, tagInvitations, activityNotes) {
     search: {
       enabled: true,
       subjectAreas: SUBJECT_AREAS,
-      localSearch: false,
+      localSearch: true,
```
melisabok: Home page should paginate the results, so the search should be remote.

venues/ICLR.cc/2019/webfield/bidWebfield.js
```diff
@@ -360,7 +360,7 @@ function renderContent(validNotes, tagInvitations, metadataNotesMap) {
       container: '#allPapers',
       search: {
         enabled: true,
-        localSearch: false,
+        localSearch: true,
```
melisabok: Bids can use local search because we load all the notes, right?

mspector: ah right. yes, makes sense

melisabok:
Local search is related to how we load the notes. With pagination -> localSearch: false, All notes -> localSearch: true

zbialecki:
I forgot about the pagination issue. I'm going to close this PR and look into why it's not picking up Subject Areas for the future.

PR #753, opened 2018-08-28
matching demo

mspector:
some matching-related updates to the scripts

also adds a throwaway script for setting up the matching demo. Let me know if you think it doesn't really belong in the repo


mspector:
@mohituniyal could you review this? it shouldn't really affect anyone else. I took out the other changes, just kept the changes to the iclr19.py file

PR #763, opened 2018-09-12
Improve formatting of empty message 

zbialecki:
and don't show sort options when table is empty


zbialecki:
I also removed some console.log statements that I didn't think should be in the live versions.

PR #764, opened 2018-09-13
allow your submitted notes tab to include papers that were submitted …

mspector:
…by the user but not part of authorids

### DIFF
--- a/venues/ICLR.cc/2019/Conference/webfield/authorWebfield.js
+++ b/venues/ICLR.cc/2019/Conference/webfield/authorWebfield.js
@@ -163,7 +163,7 @@ function load() {
     duedate: true,
     replyto: true,
     details:'replytoNote,repliedNotes'
-  }).then(result => result.invitations)
+  }).then(function(result) {return result.invitations;})
 
   return userGroupsP
   .then(function(userGroups) {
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2019/Conference/webfield/authorWebfield.js
```diff
@@ -252,6 +252,10 @@ function renderContent(notes, submittedNotes, assignedNotePairs, assignedNotes,
     if (!_.includes(COMMENT_EXCLUSION, note.invitation)) {
       commentNotes.push(note);
     }
+    if (note.invitation === SUBMISSION_ID &&
+        !_.includes(_.map(authorNotes, n=>n.id), note.id)) {
```
zbialecki: So unfortunately in Webfield code we can't use any ES6 features, so => needs to be replace with a regular function.

mspector: @zbialecki I actually see the => functions in a bunch of other places in the code - is that bad?

zbialecki: If it's in webfield files, yes.

zbialecki: It will cause JS errors in all older browsers.

venues/ICLR.cc/2019/Conference/webfield/authorWebfield.js
```diff
@@ -252,6 +252,10 @@ function renderContent(notes, submittedNotes, assignedNotePairs, assignedNotes,
...
+      authorNotes.push(note);
```
zbialecki: Also, I don't really understand what this does. It seems like you're adding a submitted note to the author notes list, but only if it already exists in the author note list??

zbialecki: Sorry, my bad. I missed the !, so you only add it if it's not in the list.

PR #771, opened 2018-09-17
Feature/test paper submissions

mspector:
@mohituniyal I wrote this to help me test the later stages of the conference, is this useful to you?

@melisabok I'm not sure how this fits into our discussion about the right way to share files. For example, should the function `resubmit` go into the file `iclr19.py`?

this script shows how we can write scripts that are also modules. you can do the following:

```
from resubmit_papers import resubmit
resubmit(client, old_invitation, new_invitation)
```


mohituniyal:
@mspector Definitely useful for preparing realistic dummy data for local testing. Thanks.

PR #774, opened 2018-09-17
Feature/groups and invitations script

mspector:
this PR does a few things:

1) it creates `invitations.py` and `groups.py`, scripts for managing per-paper invitations and groups. `invitations.py` contains some functions that can be used for workflow stage transitions. There aren't any of these functions for `groups.py` yet, but you could imagine how it could be updated to support various functionality.

2) it moves the templates out of `iclr19.py`, which seems right to me. let me know if you disagree.

@mohituniyal @melisabok, interested to hear what you think. We don't necessarily have to merge this right away, just wanted to keep brainstorming about the best way to maintain the ICLR scripts


melisabok:
It looks good to me. I like the idea of having a script to create groups and invitations. 

some questions:
- iclr19.py is not executable, right? it contains all the definitions and constants.
- Could you give different examples of how to run groups.py and invitations.py?. 
- Can we run them in different order? I guess we can. They don't depend each other.

mspector:
That’s right, iclr19.py is not executable. Does it still make sense to have all the constants in a central location like that?

I’ll write some more examples. And yes, the idea is that these scripts will be like command line utilities that can be executed whenever we need to make common changes. The scripts are also importable so that their inner functions can be used to implement the scripts that transition from one stage to another. 

What I would like to do next is to go through the workflow and define all the operations that we’ll have to do, and then define those operations in functions within the utility scripts. Then the stage transition scripts can import and call these functions.

melisabok:
if you need to share the constants between both files (groups.py and invitations.py) and I think it is ok. Otherwise you can put them in each file. 

I like the versatility of using these scripts from command line or from the stage scripts. 💯 

It would be nice to write the examples or the stage scripts so we realize what is the interface needed in each script.

mspector:
we're not done with the ICLR scripts but I think I would like to merge this to keep the PR from getting too big. is that ok with you @melisabok ?

venues/ICLR.cc/2019/Conference/python/post-submissions-stage.py
```diff
@@ -0,0 +1,64 @@
...
+    parser.add_argument('--baseurl', help="base url")
+    parser.add_argument('--username')
+    parser.add_argument('--password')
+    args = parser.parse_args()
+
+    client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+
+    client.post_invitation(iclr19.blind_submission_inv)
+
+    original_notes = openreview.tools.iterget_notes(client,invitation = iclr19.submission_inv.id)
```
melisabok: original_notes is an iterator or a list?

melisabok: nerver mind, it is an iterator! Cool

mspector:
thanks @melisabok !

PR #778, opened 2018-09-19
update author group template

mspector:


venues/ICLR.cc/2019/Conference/python/groups.py
```diff
@@ -31,12 +31,15 @@
...
-        iclr19.PROGRAM_CHAIRS_ID
+        iclr19.PROGRAM_CHAIRS_ID,
+        iclr19.PAPER_AUTHORS_TEMPLATE_STR
     ],
     'writers': [iclr19.CONFERENCE_ID],
     'signatures': [iclr19.CONFERENCE_ID],
-    'signatories': [iclr19.CONFERENCE_ID],
+    'signatories': [
+        iclr19.CONFERENCE_ID,
+        iclr19.PAPER_AUTHORS_TEMPLATE_STR
```
melisabok: signatories should be only the authors group

venues/ICLR.cc/2019/Conference/python/groups.py
```diff
@@ -128,8 +129,14 @@
 }
 
 def create_and_post(client, paper, template_key):
-    client.post_group(openreview.Group.from_json(
-        openreview.tools.fill_template(group_templates[template_key], paper)))
+    group_to_post = openreview.Group.from_json(
+        openreview.tools.fill_template(
+            group_templates[template_key], paper))
+
+    if template_key == "Paper/Authors":
```
melisabok: I would define a new parameter 'members' that is optional. And you should send the members in the post_submissions_stage.

venues/ICLR.cc/2019/Conference/python/groups.py
```diff
@@ -128,8 +129,14 @@
...
+        group_to_post.members = client.get_note(id=paper.original).content["authors"]
```
melisabok: It should be authorids instead of authors

venues/ICLR.cc/2019/Conference/python/groups.py
```diff
@@ -127,9 +125,15 @@
...
 
-def create_and_post(client, paper, template_key):
-    client.post_group(openreview.Group.from_json(
-        openreview.tools.fill_template(group_templates[template_key], paper)))
+def create_and_post(client, paper, template_key, members=[]):
+    group_to_post = openreview.Group.from_json(
+        openreview.tools.fill_template(
+            group_templates[template_key], paper))
+
+    if members:
```
melisabok: could you pass the members to the template so you need to add the if?

mspector: just to clarify, are you just asking for us to remove the if statement, or are you asking us to add the members to the template object before turning it into a group?

melisabok: I thought that you could pass the members in the fill_template function but I see that is general function.

PR #779, opened 2018-09-20
Fix bug where when AC has no assigned papers

zbialecki:


mohituniyal:
@zbialecki this does not seem to work. Still getting this error:
screen shot 2018-09-20 at 1 28 49 am

zbialecki:
OK I think I fixed it for real. @mohituniyal can you test as well?

mohituniyal:
> OK I think I fixed it for real. @mohituniyal can you test as well?

It is fixed now. Thanks @zbialecki . Merging this.

PR #782, opened 2018-09-20
Disable search for no bid tab

zbialecki:


zbialecki:
Should be good to merge now

PR #783, opened 2018-09-20
Review assignment stage

mspector:
does *not* include a script to deploy the assignments. that will be separate

### DIFF
--- a/venues/ICLR.cc/2019/Conference/python/invitations.py
+++ b/venues/ICLR.cc/2019/Conference/python/invitations.py
@@ -246,20 +246,33 @@ invitation_templates = {
 
 current_timestamp = lambda: int(round(time.time() * 1000))
 
-def enable_invitation(template_key, paper):
-    new_invitation = openreview.Invitation.from_json(
-        openreview.tools.fill_template(invitation_templates[template_key], paper))
+def enable_invitation(template_key, target=None):
+    if target:
+        new_invitation = openreview.Invitation.from_json(
+            openreview.tools.fill_template(invitation_templates[template_key], target))
+    else:
+        new_invitation = openreview.Invitation.from_json(
+            invitation_templates[template_key], target)
+
     return new_invitation
 
-def disable_invitation(template_key, paper):
-    new_invitation = openreview.Invitation.from_json(
-        openreview.tools.fill_template(invitation_templates[template_key], paper))
+def disable_invitation(template_key, target=None):
+    if target:
+        new_invitation = openreview.Invitation.from_json(
+            openreview.tools.fill_template(invitation_templates[template_key], target))
+    else:
+        new_invitation = openreview.Invitation.from_json(
+            invitation_templates[template_key])
+
     new_invitation.expdate = current_timestamp()
     return new_invitation
 
 def enable_and_post(client, paper, template_key):
-    new_inv = enable_invitation(template_key, paper)
+    new_inv = enable_invitation(template_key, target=paper)
     return client.post_invitation(new_inv)
+
+def disable_bids(client):
+    return client.post_invitation(disable_invitation('Add_Bid'))
 
 if __name__ == '__main__':
     ## Argument handling
@@ -279,7 +292,7 @@ if __name__ == '__main__':
         for template in args.invitations:
             assert template in invitation_templates, 'invitation template not defined'
             if args.disable:
-                new_invitation = disable_invitation(template, paper)
+                new_invitation = disable_invitation(template, target=paper)
             else:
                 new_invitation = enable_invitation(template, paper)
             posted_invitation = client.post_invitation(new_invitation)
--- a/venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
+++ b/venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
@@ -1,0 +1,103 @@
+'''
+Reviewer Assignment Stage:
+- Bidding Task / interface is closed.
+- Initial paper-reviewer and paper-areachair matches are computed.
+- Program chairs browse match using OpenReview Matching System and make hand-adjustments as needed.
+- Paper-reviewer and paper-areachair assignments are finalized and deployed.
+
+'''
+
+import argparse
+import openreview
+import iclr19
+import invitations
+import notes
+import time
+import json
+
+import matcher
+
+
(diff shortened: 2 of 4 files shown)


venues/ICLR.cc/2019/Conference/python/invitations.py
```diff
@@ -246,21 +246,34 @@
 
 current_timestamp = lambda: int(round(time.time() * 1000))
 
-def enable_invitation(template_key, paper):
-    new_invitation = openreview.Invitation.from_json(
-        openreview.tools.fill_template(invitation_templates[template_key], paper))
+def enable_invitation(template_key, target=None):
```
melisabok: What does target mean?

venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
```diff
@@ -0,0 +1,103 @@
...
+    assert all(['~' in member for member in reviewers_group.members]), 'not all reviewers have been converted to profile IDs'
+    reviewer_profiles = client.get_profiles(reviewers_group.members)
+
+    invitations.disable_bids(client)
+
+    # create metadata
+    metadata_notes = []
+    metadata_inv = client.post_invitation(iclr19.metadata_inv)
+    for blind_note in blind_submissions:
+        original_note = client.get_note(id=blind_note.original)
```
melisabok: Instead of doing this you can ask for the original note and tags in  the get blinded notes. 

You need to add the parameter ```details=original,tags``` and then you will get in each note:

```
{
id: 1111,
....
details: {
   original: {
      id: 2222
   } ,
   tags: [
  
   ]
}
}
```

melisabok: doing 2 requests per blinded note is going to be too slow.

venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
```diff
@@ -0,0 +1,103 @@
...
+        paper_bids = client.get_tags(invitation=iclr19.add_bid.id)
```
melisabok: for each blinded note are you doing the same call?

venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
```diff
@@ -0,0 +1,103 @@
...
+    demands = [config['max_users']] * len(metadata_notes)
+
+    # run initial match
+    encoder = matcher.metadata.Encoder(
+        metadata_notes,
+        config,
+        [r.id for r in reviewer_profiles])
+
+    flow_solver = matcher.Solver(supplies, demands, encoder.cost_matrix, encoder.constraint_matrix)
+    solution = flow_solver.solve()
```
melisabok: I don't know why you are doing the solver here, it is just for the demo?

venues/ICLR.cc/2019/Conference/python/invitations.py
mspector: I want the invitations script to be able to modify templated invitations (per-paper, like official review) *or* non-templated invitations (like add bid). the "target" is the paper to template against. if it's None, it means that there is nothing to template against.

if you don't like the variable name, I don't mind changing it. but that's its purpose.

venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
mspector: oops... that is a mistake

venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
mspector: no, we need this to initialize the matching browser based on TPMS scores and bid scores. we don't have scores right now, so there is dummy data. but this script is supposed to generate assignments with the solver.

mspector: I think I understand what you're saying now - are you wondering why we don't do the matching in a separate script?

I'm not sure. maybe we should. let's discuss it

melisabok: This code is in the matching repo, why don’t you run the matching from there? 

You have to create only the metadata notes here, right?

mspector:
@melisabok I addressed your comments, I think it's ready now, let me know what you think when you have a chance

melisabok:
What do you mean? Regenerate the metadata notes?

mspector:
yeah, the matching code assumes just one metadata note per paper and can get screwed up if multiple are posted

melisabok:
Well, we should support the regeneration of the metadata notes in the future. We are going to need it. Is it possible to reuse some code from the ECCV or UAI conf?

mspector:
@melisabok @mohituniyal I just made two changes: 

1) the previously existing metadata, assignment notes, and config are all deleted before anything else happens. this lets you re-run the script.

2) I took out the assertion and added a print line that warns the user when not all reviewer group members were converted to profile IDs. the users without profiles are ignored in the metadata creation, so now no outside step is required when running this step

venues/ICLR.cc/2019/Conference/python/review-assignment-stage.py
```diff
@@ -0,0 +1,61 @@
...
+import invitations
+import notes
+import time
+import json
+
+import matcher
+
+def clear(client, invitation):
+    for note in openreview.tools.iterget_notes(client, invitation = invitation):
+        client.delete_note(note)
```
melisabok: Does delete the note from the db?

mspector: I'm not sure - from the client library, it looks like it's sending a "delete" request to the backend. does that delete it from the db or set a ddate? it may delete it from the db.

would that be a problem?

mspector: here's the code: https://github.com/iesl/openreview-py/blob/master/openreview/openreview.py#L493

melisabok: It removes the notes from the database, I think it is better than setting a ddate. Because metadata notes are heavy and we don't want to keep them. 

I wonder if for some reason you need to keep them.

PR #785, opened 2018-09-21
Adding AC webfield changes to bidding stage AC webfield

mohituniyal:
Going forward we need to make sure that the issues we observe with AC webfield in a particular stage may be present in other stage's AC webfields.

### DIFF
--- a/venues/ICLR.cc/2019/Conference/webfield/areachairWebfieldBiddingEnabled.js
+++ b/venues/ICLR.cc/2019/Conference/webfield/areachairWebfieldBiddingEnabled.js
@@ -235,17 +235,16 @@ var renderHeader = function() {
 
 var renderStatusTable = function(profiles, notes, completedReviews, metaReviews, reviewerIds, container) {
   var rows = _.map(notes, function(note) {
-    var revIds = reviewerIds[note.number];
+    var revIds = reviewerIds[note.number] || Object.create(null);
     for (var revNumber in revIds) {
       var uId = revIds[revNumber];
       revIds[revNumber] = _.get(profiles, uId, { id: uId, name: '', email: uId });
     }
 
     var metaReview = _.find(metaReviews, ['invitation', CONFERENCE + '/-/Paper' + note.number + '/Meta_Review']);
+    var noteCompletedReviews = completedReviews[note.number] || Object.create(null);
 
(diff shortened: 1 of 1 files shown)


mspector:
great point. yet another reason why we should start abstracting more of the webfield functionality

venues/ICLR.cc/2019/Conference/webfield/areachairWebfieldBiddingEnabled.js
```diff
@@ -235,17 +235,16 @@ var renderHeader = function() {
 
 var renderStatusTable = function(profiles, notes, completedReviews, metaReviews, reviewerIds, container) {
   var rows = _.map(notes, function(note) {
-    var revIds = reviewerIds[note.number];
+    var revIds = reviewerIds[note.number] || Object.create(null);
```
melisabok: What does it mean ```Object.create(null)```?

melisabok: You can use just ```{}```

mohituniyal: This code is copied into this file from https://github.com/iesl/openreview-scripts/pull/779 .
Maybe @zbialecki can better comment on this

melisabok: We should try to avoid the duplication of code. I understand that you need to keep different versions of the webfield, one per stage. The downside is that is difficult to maintain.

zbialecki: I agree with Melisa, maintaining 2 versions of this file that we have to keep in sync is not a long term solution.

Object.create(null) is used to create an map that doesn't have all the builtin properties of an object. So using `prop in object` is safe, and calling `.hasOwnProperty()` is not necessary. So it's better if you're going to be using `for...in` loops.

melisabok: I imagined that. Thanks for the explanation. 

Webfields are becoming too complex, we need to work on this after ICLR.

PR #788, opened 2018-09-21
Updated reviewer questionnaire template

mohituniyal:


venues/ICLR.cc/2019/Conference/python/iclr19.py
```diff
@@ -336,21 +336,16 @@
         'writers': {'values-regex': '~.*'},
         'signatures': {'values-regex': '~.*'},
         'content': {
-            'title': {
```
melisabok: All the notes need a title, otherwise we will render a 'No Title' string is that ok?

mspector: mohit and I talked about this, it seems to be ok

PR #794, opened 2018-09-21
fix get author console and improve request calls

melisabok:
- remove get tag invitations
- get the group by web=true
- remove get notes in the first stage
- get notes by authorids to decide the author console rendering
- enable spinner
- remove commented code


zbialecki:
Looks good to me

PR #802, opened 2018-09-24
withdraw submission

mspector:
beginning of submission withdrawals. process function doesn't do everything it's supposed to yet, but I think this should be merged for now (@mohituniyal @melisabok let me know if you disagree)


mspector:
@melisabok sounds good, don't merge this yet, I can fix the process function to do the right thing now (I was putting it off because of the issue we discussed related with the blind note references, but now it's resolved).

The problem with the single invitation is that only the author group of the specific paper should have permission to withdraw

mspector:
@melisabok @mohituniyal this should now be ready. notice that the post-submissions-stage.py file now enabled the withdrawal invitation

mspector:
@melisabok @mohituniyal I think this is ready for now

PR #803, opened 2018-09-24
Rebuttal stage

mspector:
requires https://github.com/iesl/openreview-scripts/pull/802/files

### DIFF
--- a/venues/ICLR.cc/2019/Conference/python/invitations.py
+++ b/venues/ICLR.cc/2019/Conference/python/invitations.py
@@ -76,23 +76,6 @@ revise_review_template = {
     }
 }
 
-add_revision_template = {
-    'id': iclr19.CONFERENCE_ID + '/-/Paper<number>/Add_Revision',
-    'readers': ['everyone'],
-    'writers': [iclr19.CONFERENCE_ID],
-    'invitees': [iclr19.CONFERENCE_ID + '/Paper<number>/Authors'],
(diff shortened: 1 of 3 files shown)


mspector:
@melisabok @mohituniyal this is also ready

venues/ICLR.cc/2019/Conference/python/invitations.py
```diff
@@ -56,6 +56,26 @@
 with open(os.path.abspath('../process/officialReviewProcess.js')) as f:
     official_review_template['process'] = f.read()
 
+revise_review_template = {
+    'id': iclr19.CONFERENCE_ID + '/-/<reviewer_id>/Revise/Review',
```
mohituniyal: @mspector Just wondering why we are using "Revise/Review" instead of "Revise_Review", unlike "Add_Review", "Add_Bid", etc.

mspector: great question. when an invitation ID is rendered, the rule that we currently have in place takes the last two segments of the ID, when split by the "/" character. 

for example, `ICLR.cc/2019/Conference/-/Paper123/Official_Review` becomes `Paper123 Official Review`

In this case, I want to only display the words `Revise Review`. So that means I need to format the string as `.../Revise/Review`.

It's not great. We really need to come up with a better convention. In my opinion, it should render only the last component, and our other invitations should be renamed. But we haven't gotten around to this.

PR #809, opened 2018-09-26
allow TPMS sorting

mspector:


venues/ICLR.cc/2019/Conference/python/bidding_stage.py
```diff
@@ -27,13 +28,58 @@
...
+    user_score_notes = {}
+    for paper in openreview.tools.iterget_notes(client, invitation=iclr19.BLIND_SUBMISSION_ID):
+        for user_id in valid_reviewer_ids + valid_ac_ids:
+            if user_id not in user_score_notes:
+                user_score_notes[user_id] = openreview.Note.from_json({
+                    'invitation': iclr19.SCORES_INV_ID,
+                    'readers': [user_id],
+                    'writers': ['ICLR.cc/2019/Conference'],
+                    'content': {
+                        'users': [user_id],
```
melisabok: why this is an array and not an string?

mspector: ah, it was left over from the way it was designed before. I will change it

venues/ICLR.cc/2019/Conference/webfield/bidWebfield.js
```diff
@@ -48,15 +48,28 @@ function load() {
     });
   });
 
-  var metadataNotesP = Webfield.getAll('/notes', {invitation: METADATA_INVITATION_ID}).then(function(allNotes) {
-    var metadataMap = {};
-    for (var i = 0; i < allNotes.length; i++) {
-      metadataMap[allNotes[i].forum] = allNotes[i].content.groups;
+  var userScoresP = Webfield.getAll('/notes', {invitation: USER_SCORES_INVITATION_ID}).then(function(scoreNotes) {
```
melisabok: Again we are getting all the notes for all the users. We need to get a single note, one option is to filter by content.user == profile.id and I can add an index in mongo. 

```
GET /notes?invitation=[invitaiton_id]&content.user=~profile_id
```

but content.user should an string.

venues/ICLR.cc/2019/Conference/python/bidding_stage.py
melisabok: If we use an string we can filter the notes by content.user, see my comment below.

venues/ICLR.cc/2019/Conference/python/bidding_stage.py
```diff
@@ -27,13 +28,58 @@
...
+                        'scores': []
+                    },
+                    'signatures': ['ICLR.cc/2019/Conference'],
+                    'forum': None,
+                    'multiReply': False
+                })
+            score_note = user_score_notes[user_id]
+
+            score_entry = {
```
melisabok: Is there a way to do this less verbose? I'm worried about having 3000 entries.

venues/ICLR.cc/2019/Conference/webfield/bidWebfield.js
mspector: I'm not sure if this helps, but the readership permissions are set so that each reviewer or AC should only receive back one note from this query. does that make it more efficient? I'm not sure

venues/ICLR.cc/2019/Conference/python/bidding_stage.py
mspector: I'm not sure - in each user's score note, we need the forum ID and score for each paper.

the most compressed way that I can think to do it is to store it as a big string, something formatted like this:

"forum=ABCxyz;score=0.76|forum=123abc;score=0.35|..."

then we could parse it in the webfield. would this be any better, though? I don't really know

PR #810, opened 2018-09-26
Fix header of reviewers console

zbialecki:


melisabok:
@zbialecki , there is a conflict in the homepage file. Could you resolve it? and then I can merge it.

zbialecki:
OK, fixed!

PR #813, opened 2018-09-26
Fix/reviewer console

melisabok:
- Change the order of the tabs.
- Check if noteNumbers is empty then don't do the request calls.
- Change limit value 2000 to 1000, max available.
- Fix get tag invitations.
- Removed console logs.
- Removed commented code.
- Apply the changes to the other webfields (I copied and pasted from line 29 to the end of the file). The difference between the 3 files are: INSTRUCTIONS and SCHEDULE_HTML variables.

### DIFF
--- a/venues/ICLR.cc/2019/Conference/webfield/reviewerWebfield.js
+++ b/venues/ICLR.cc/2019/Conference/webfield/reviewerWebfield.js
@@ -68,7 +68,6 @@ var getBlindedNotes = function(noteNumbers) {
     .then(function(result) {
       return result.notes;
     });
-
 };
 
 var getAllRatings = function(callback) {
@@ -116,7 +115,6 @@ var getReviewRatings = function(noteNumbers) {
   });
 
   return dfd.promise();
-
 };
 
 var getReviewerGroups = function(noteNumbers) {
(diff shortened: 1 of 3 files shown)


venues/ICLR.cc/2019/Conference/webfield/reviewerWebfield.js
```diff
@@ -58,12 +58,17 @@ var getPaperNumbersfromGroups = function(groups) {
...
-  var noteNumbersStr = noteNumbers.join(',');
+  if (noteNumbers.length) {
+    var noteNumbersStr = noteNumbers.join(',');
+
+    return $.getJSON('notes', { invitation: CONFERENCE + '/-/Blind_Submission', number: noteNumbersStr, noDetails: true })
+      .then(function(result) {
+        return result.notes;
+      });
+  } else {
+    return $.Deferred().resolve([]);
```
zbialecki: In general I prefer the style of "returning early" meaning that instead of doing `if (noteNumbers.length)` you would do `if (!noteNumbers.length) return ...` and then leave the rest of the function the same. This has the advantage of reducing the indentation level, making the code easier to read, and reducing the size of the diffs.

PR #814, opened 2018-09-26
return empty promises when there are no assigned papers

melisabok:


venues/ICLR.cc/2019/Conference/webfield/areachairWebfield.js
```diff
@@ -74,21 +74,29 @@ var buildNoteMap = function(noteNumbers) {
...
+  var blindedNotesP;
+  var metaReviewsP;
 
-  var blindedNotesP = Webfield.get('/notes', {
-    invitation: BLIND_SUBMISSION_ID, number: noteNumbersStr, noDetails: true
-  })
-  .then(function(result) {
-    return result.notes;
-  });
+  if (noteNumbers.length) {
```
zbialecki: As I mentioned in my other review I think this would be better as `if (!noteNumbers.length) { return $.Deferred().resolve([]); }`

venues/ICLR.cc/2019/Conference/webfield/areachairWebfieldReviewerAssignmentStage.js
```diff
@@ -194,7 +212,7 @@ var getUserProfiles = function(userIds) {
     _.forEach(result.profiles, function(profile) {
       var name = _.find(profile.content.names, ['preferred', true]) || _.first(profile.content.names);
       profile.name = _.isEmpty(name) ? view.prettyId(profile.id) : name.first + ' ' + name.last;
-      profile.email = profile.content.preferredEmail;
+      profile.email = profile.content.preferredEmail || profile.content.emails[0];
```
melisabok: I think this change was in the first file and not here?

venues/ICLR.cc/2019/Conference/webfield/areachairWebfieldReviewerAssignmentStage.js
```diff
@@ -248,7 +266,7 @@ var renderStatusTable = function(profiles, notes, completedReviews, metaReviews,
   var order = 'desc';
   var sortOptions = {
     Paper_Number: function(row) { return row[1].number; },
-    Paper_Title: function(row) { return _.toLower(row[2].content.title); },
+    Paper_Title: function(row) { return _.toLower(_.trim(row[2].content.title)); },
```
melisabok: same here

venues/ICLR.cc/2019/Conference/webfield/areachairWebfield.js
melisabok: It should be ready now

venues/ICLR.cc/2019/Conference/webfield/areachairWebfieldReviewerAssignmentStage.js
zbialecki: Yeah, it looks like it.

venues/ICLR.cc/2019/Conference/webfield/areachairWebfield.js
mspector: looks like you are still doing "return last", is that ok? seems ok to me since (I think) it's just a style thing, so I'm going to merge it

zbialecki: Yeah, I chose a bad example to comment on because in this case the code isn't actually returning, it's setting variables so both the if and the else are needed.

PR #817, opened 2018-09-27
Iclr19/split post submission

mspector:
splits the post-submission-stage script into two parts.

part 1 does only two things: removes the writers from the original notes, and replaces the ICLR homepage webfield with the post-submission version.

part 2 is identical to the original script.

I changed the post-submission webfield because I was worried that users would see no papers in the "all submissions tab" and panic. But we don't necessarily have to do this if people object.


melisabok:
You could update the homepage after the blinded notes are created.

mspector:
@melisabok yes, I could do that- I would need to create a separate webfield file though, right? I thought this would be better, so we don't have to keep more files in sync. But I can do it either way.

PR #822, opened 2018-09-28
Updated public-comment invitation to allow anonymous signatures

mohituniyal:


mohituniyal:
https://github.com/iesl/openreview/pull/1148 should work for this PR to take effect.

PR #832, opened 2018-10-05
Added a conference level invitation for withdrawn invitation

mohituniyal:


venues/ICLR.cc/2019/Conference/python/iclr19.py
```diff
@@ -571,10 +572,36 @@
...
+        },
+        'signatures': {
+            'description': 'How your identity will be displayed with the above content.',
+            'values': [CONFERENCE_ID]
+        },
+        'writers': {
+            'description': 'Users that may modify this record.',
+            'values':  [CONFERENCE_ID]
+        },
+        'content': {}
```
mspector: this isn't a problem in my opinion, but just FYI: leaving the content field empty will allow you to pass in any content fields that you want, free of restriction. it doesn't mean that you can't pass in any content (which is what you might think, intuitively)

venues/ICLR.cc/2019/Conference/python/iclr19.py
```diff
@@ -571,10 +572,36 @@
 
 })
 
-
+# This is the Conference level Invitation for all withdrawn submissions
+withdrawn_submission_invitation = openreview.Invitation.from_json({
+    'id': WITHDRAWN_SUBMISSION_TEMPLATE_STR,
```
mspector: I don't think this needs to be a templated invitation. I think you can have it be a global invitation like the blind submission

venues/ICLR.cc/2019/Conference/python/iclr19.py
melisabok: It is the same invitation of last year. I said that is ok because we are not going to use the UI or another signature to create notes

PR #834, opened 2018-10-05
Updated disable_bids to empty invitees for Add_Bids invitation

mohituniyal:
PR contains changes to empty the invitees field of the Add_Bid invitation.

Note: The way add_bid is getting enabled or disabled right now is the same enable/disable function runs for each blind submission, which is not at all required. This will be fixed in another PR.


mohituniyal:
I found a problem with the change. Will create the PR again.

PR #852, opened 2018-10-19
Update webfields to be more compatible with tab history update

zbialecki:
Related to https://github.com/iesl/openreview/pull/1185


zbialecki:
The most important webfields to update after this change is merged is the ICLR homepage, as that was where people were complaining about not being directed back to the tab they were viewing.

PR #854, opened 2018-10-24
Akbc/archival status update

mohituniyal:
1. Fixed archival status update - file: python/invitations.py
2. fixed a small bug on AKBC homepage, where if user does not have a your-consoles page, the homepage was showing blank tabs after submission.
3. Generalized all python scripts so that we do not have to make a lot of changes for further conferences.

### DIFF
--- a/venues/AKBC.ws/2019/Conference/python/assign.py
+++ b/venues/AKBC.ws/2019/Conference/python/assign.py
@@ -1,7 +1,7 @@
 
 
 import openreview
-import akbc19 as conferenceConfig
+import akbc19 as conference_config
 import notes
 import groups
 import invitations
@@ -23,7 +23,7 @@ if __name__ == '__main__':
     client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
 
     assignment_notes = openreview.tools.iterget_notes(client,
-        invitation=conferenceConfig.ASSIGNMENT_INV_ID,
+        invitation=conference_config.ASSIGNMENT_INV_ID,
         details='forumContent')
 
     def get_number_from_details(note_details):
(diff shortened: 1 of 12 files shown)


venues/AKBC.ws/2019/Conference/python/invitations.py
```diff
@@ -16,49 +16,20 @@
...
-        'value-dropdown': {},
-        'description': 'Update archival status.',
-        'required': True
-    }
-}
-
-archival_status_template = {
-    'id': conferenceConfig.ARCHIVAL_STATUS_TEMPLATE_STR,
+change_archival_status_template = {
+    'id': conferenceConfig.CONFERENCE_ID + '/-/Paper<number>/Change_Archival_Status',
```
mspector: I'm interested in other opinions, but my thought about naming invitations is that they should be named for nouns, not verbs. So instead of "Change_Archival_Status", maybe it should just be "Archival_Status"?

@melisabok what do you think?

venues/AKBC.ws/2019/Conference/python/assign.py
```diff
@@ -1,7 +1,7 @@
 
 
 import openreview
-import akbc19
+import akbc19 as conferenceConfig
```
mspector: really minor thing (which you don't have to follow if you don't want): the conventional style for python variables is to use snake_case instead of camelCase

venues/AKBC.ws/2019/Conference/python/invitations.py
melisabok: I like using nouns, 'Archival Status' is better. 

I know that we have 'Add Revision', we should rename it in the future.

mspector: ah yeah, it should probably just be "Revision"

mohituniyal: good point, but I feel that a user would find "Add Revision" and "Change Archival Status" more instructive than "Revision" and "Archival Status". What do you guys think ?

venues/AKBC.ws/2019/Conference/python/assign.py
mohituniyal: Thanks for this suggestion. I have made changes in accordance.

PR #856, opened 2018-10-24
Fix/webfield controller update

zbialecki:
Makes all webfield code compatible with https://github.com/iesl/openreview/pull/1175

The main changes are:
- No references to `window` or `document` allowed
- No HTML webfields allowed (still working on this)


venues/ICLR.cc/ICLR-landing-webfield.js
```diff
@@ -0,0 +1,23 @@
...
+  { url: '/group?id=ICLR.cc/2016/workshop', name: 'ICLR 2016 Workshop Track' },
+  { url: '/group?id=ICLR.cc/2014', name: 'ICLR 2014 Workshop Track' },
+  { url: '/group?id=ICLR.cc/2013', name: 'ICLR 2013 Conference Track' }
+];
+
+Webfield.ui.setup('#group-container', GROUP_ID);
+
+Webfield.ui.header(HEADER.title, HEADER.description, { underline: true });
+
+Webfield.ui.linksList(VENUE_LINKS);
```
melisabok: If I had this linksList method I would start to migrate these kind of webfields in live.

zbialecki: I could make a separate PR with just the webfield.ui.linksList function

melisabok:
Ok, I updated most of them. There are some webfields that we are not currently using like UAI console and I will leave them.

PR #858, opened 2018-10-26
ACM SIGIR Badging scripts

mohituniyal:


venues/ACM.org/SIGIR/Badging/python/acm19.py
```diff
@@ -0,0 +1,260 @@
...
+    'pdf': {
+        'fieldDisplayLabel': 'PDF',
+        'description': 'Either upload a PDF file or provide a direct link to your artifact (link must begin with http(s))',
+        'order': 7,
+        'value-regex': 'upload|(http|https):\/\/.+',
+        'required':True
+    }
+}
+
+class ArtifactSubmission(openreview.Invitation):
```
melisabok: you can use openreview. Invitation instead, no need to create a class that is being used once.

mohituniyal: Resolved.

PR #859, opened 2018-10-29
Disable confirmation message for single reminder emails 

zbialecki:
And immediately show last sent date.

Any help testing this would be appreciated since I don't have any paper assignment groups set up.


melisabok:
I will test it

zbialecki:
I resolved Mohit's first 2 issues. For now I am not going to automatically add the "Last sent" text when multiple reminders are sent to reviewers via the modal dialog.

PR #860, opened 2018-10-30
Fix/note action buttons

zbialecki:
Enable action buttons in activity tab in ICLR


melisabok:
I don't think we need to enable the action buttons here, I recommend to close this PR.

PR #866, opened 2018-11-02
Added "html" field to ACM submission invitation

mohituniyal:


venues/ACM.org/SIGIR/Badging/python/acm19.py
```diff
@@ -156,12 +156,18 @@
                 ],
                 'required': True
             },
+            'html': {
+                'fieldDisplayLabel': 'Html',
+                'description': 'Either provide a direct url to your artifact (link must begin with http(s)) or upload a PDF file',
+                'order': 8,
```
melisabok: Could you put another order number? this is the order how the fields are rendered in the form.

mohituniyal: Thanks for noticing this. It's fixed.

PR #869, opened 2018-11-07
Fix/iclr areachair reminder links

zbialecki:
Fix link sent when a single reminder email is triggered


melisabok:
I found another issues related to the reminder: sendgrid doesn't send the message content in HTML, it is using plain text. We need to send the url without the ```<a>``` tag.

PR #879, opened 2018-11-16
Colt19/program committee placeholder webfield

mspector:


melisabok:
I added the init.py to setup the conference using the builder.

PR #882, opened 2018-11-19
filter out author's own papers in bidding interface

mspector:


venues/AKBC.ws/2019/Conference/webfield/bidWebfield.js
```diff
@@ -86,13 +89,17 @@ function load() {
...
+  return $.when(notesP, authoredNotesP, tagInvitationsP, userScoresP);
 }
 
 
 // Display the bid interface populated with loaded data
-function renderContent(validNotes, tagInvitations, metadataNotesMap) {
-
+function renderContent(validNotes, authoredNotes, tagInvitations, metadataNotesMap) {
+  validNotes = _.filter(validNotes, function(note){
+    return !_.includes(_.map(authoredNotes, function(note){
```
zbialecki: I would move the _.map outside the _.filter to it's own variable, so it's not being re-computed every iteration

PR #884, opened 2018-11-19
Fix/midl builder

melisabok:


venues/MIDL.io/2019/Conference/python/recruit_area_chairs.py
```diff
@@ -0,0 +1,82 @@
...
+
+10. The program chairs will collect all the AC recommendations and will do a final check on the consistency of the reviewer scores + AC scores + AC recommendations. The final decisions will be announced on February 20th, 2019.
+
+In summary, finally, the Program chairs will go through the 4 scores (3 Reviewers + 1 AC review + AC recommendation) to detect any major inconsistencies, if any. In most cases, AC recommendation will be the base resulting decision, however, we will check outlier cases, where for instance 3 reviewers return with negative results and the AC returns with a positive decision and such.
+
+'''
+
+conference.recruit_reviewers(emails = formatted_acs, title = title, message = message, reviewers_name = 'Area_Chairs')
```
melisabok: to remind reviewers/areachairs you can add the parameter: remind = True:

```
conference.recruit_reviewers(title = title, message = message, reviewers_name = 'Area_Chairs', remind = True)
```

melisabok: this will send an email with the same text and subject: 'Reminder' + title to the invited reviewers that haven't responded yet.

venues/MIDL.io/2019/Conference/python/recruit_area_chairs.py
```diff
@@ -0,0 +1,82 @@
...
+builder.set_homepage_header({
+'title': 'Medical Imaging with Deep Learning',
+'subtitle': 'MIDL 2019 Conference',
+'deadline': 'Submission Deadline: 13th of December, 2018',
+'date': '8-10 July 2019',
+'website': 'http://2019.midl.io',
+'location': 'London',
+'instructions': 'Full papers contain well-validated applications or methodological developments of deep learning algorithms in medical imaging. There is no strict limit on paper length. However, we strongly recommend keeping full papers at 8 pages (excluding references and acknowledgements). An appendix section can be added if needed with additional details but must be compiled into a single pdf. The appropriateness of using pages over the recommended page length will be judged by reviewers. All accepted papers will be presented as posters with a selection of these papers will also be invited for oral presentation.'
+})
+builder.set_conference_type(openreview.builder.SingleBlindConferenceType)
```
pmandler: Do you need to set the homepage_header to be able to use the conference_builder to send the recruiting emails?

melisabok: Yes, the invitation response web page it show this header.

pmandler:
I pushed some code onto this branch when I should have created a new branch.  I don't think there is an easy way to undo that.
I have some questions about the submissionProcess code that I've added.
1) The comment invitation seems to only be visible to people that have logged it. It should be visible to everyone, and just not able to respond until logged in.  I thought setting the readers to 'everyone' was sufficient.  Am I missing something?
2) I have a bunch of promises some of which need to be run in order.  Can I use Promise.all()?
3) The Comment invitation has no process function.  Since comments are only visible by paper authors, PCs and ACs I don't want to send emails if a comment that is written by someone else receives a response. Not sure when emails are warranted. The way they have specified it, if a reviewer comments and an author responds, the reviewer can not see the response.

pmandler:
Changing the invitees and the signatures works.

PR #887, opened 2018-12-05
Fix/midl setup

melisabok:


venues/MIDL.io/2019/Conference/python/recruit_reviewers.py
```diff
@@ -0,0 +1,84 @@
...
+Ipek Oguz, Gozde Unal and Ender Konukoglu
+Program Chairs for Medical Imaging with Deep Learning 2019
+
+'''
+
+group = client.get_group(id = conference.get_reviewers_id())
+
+group.writers = [conference.id, conference.get_program_chairs_id(), conference.get_area_chairs_id()]
+
+formatted_members = [m.lower() if '@' in m else m for m in group.members]
```
melisabok: I'm setting the members to lowercase now but I think we need to replace the emails with profile ids.

PR #889, opened 2018-12-05
Feature/midl reviews

pmandler:
Create review invitations.


venues/MIDL.io/2019/Conference/python/create-review-invitations.py
```diff
@@ -0,0 +1,121 @@
...
+print("Connecting to "+client.baseurl)
+
+builder = openreview.conference.ConferenceBuilder(client)
+builder.set_conference_id(config.CONFERENCE_ID)
+builder.set_conference_name('Medical Imaging with Deep Learning')
+builder.set_conference_submission_name('Full_Submission')
+conference = builder.get_result()
+conference.set_authors()
+
+submissions = client.get_notes(invitation=config.CONFERENCE_ID+'/-/Full_Submission')
```
melisabok: I would use the iterator here

venues/MIDL.io/2019/Conference/python/create-review-invitations.py
```diff
@@ -0,0 +1,121 @@
...
+    paper_num = str(paper.number)
+    paperinv = config.CONFERENCE_ID + '/-/Paper' + paper_num
+    paperGroup = config.CONFERENCE_ID + '/Paper' + paper_num
+    print(paper_num)
+    ## Reviewer group - people that can see the review invitation
+    reviewerGroup = paperGroup + '/Reviewers'
+    try:
+        client.get_group(id=reviewerGroup)
+        print("Found " + reviewerGroup)
+    except openreview.OpenReviewException as e:
```
melisabok: in tools there is get_group that returns None if the group doesn't exist.

venues/MIDL.io/2019/Conference/python/create-review-invitations.py
```diff
@@ -0,0 +1,121 @@
...
+        # is group not found, then make groups and invites
+        print(e.args[0][0])
+        if e.args[0][0].startswith('Group Not Found'):
+            client.post_group(openreview.Group(
```
melisabok: I think these groups are created when we run tools.assign, right? you don't need to create them now.

venues/MIDL.io/2019/Conference/python/create-review-invitations.py
```diff
@@ -0,0 +1,121 @@
...
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+## Initialize the client library with username and password
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print("Connecting to "+client.baseurl)
+
+builder = openreview.conference.ConferenceBuilder(client)
+builder.set_conference_id(config.CONFERENCE_ID)
```
melisabok: I guess you need to set the headers too otherwise the home page is overwritten with the default data.

pmandler: Then we should move the header info into the config.py file so we don't have to update it in a bunch of places.

melisabok: That is a good idea, may be the config file should have a function to get the conference instance.

venues/MIDL.io/2019/Conference/python/create-review-invitations.py
pmandler: It does, but it doesn't include the AC for the paper as a reader.

venues/MIDL.io/2019/Conference/python/create-review-invitations.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+    ## Confidential Review
+    review_reply = {
+        'forum': paper.id,
+        'replyto': paper.id,
+        'writers': {'values-regex': '~.*|'+paperGroup + '/AnonReviewer[0-9]+'},
+        'signatures': {'values-regex': '~.*|'+paperGroup + '/AnonReviewer[0-9]+'},
+        'readers': {
+            'values': [config.CONFERENCE_ID, conference.get_program_chairs_id(), paperGroup+'/AreaChairs'],
```
melisabok: you can user: conference.get_id() and conference.get_area_chairs_id()

venues/MIDL.io/2019/Conference/python/create-review-invitations.py
melisabok: conference.get_submission_id()

PR #903, opened 2018-12-12
implement post submissions stage using the builder

melisabok:
@pmandler we can discuss the comment process function tomorrow. 

These 2 scripts are following the conventions of ICLR and AKBC but using the builder.


venues/MIDL.io/2019/Conference/python/post-submission-stage-2.py
```diff
@@ -0,0 +1,37 @@
...
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print('connecting to {0}'.format(client.baseurl))
+
+conference = config.get_conference(client)
+
+print('open comment invitations')
+conference.set_authors()
```
mspector: this looks good to me. the API is looking nice. one comment here is that it's not obvious why the conference needs to set the authors, or what set_authors() does.

(btw, I understand why this is needed, but I'm just making a general comment about the intuitiveness of the API. I don't think it needs to change right now necessarily)

would `create_authors` make sense?

melisabok: May be we don't need to expose this method and create the authors when they are needed:
- When there is at least one invitation that contains a anonynous author group?

I'm not sure.

melisabok: It is odd to have to all this method from outside of the builder.

melisabok: That is a good idea, but I'm not sure if we need to create the author groups in all the conferences. 

For example: for an open conference with public comments we shouldn't create these groups because they are not needed. 

So I want to delay the creation of these groups until we really need them.

venues/MIDL.io/2019/Conference/python/post-submission-stage-1.py
```diff
@@ -0,0 +1,29 @@
...
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print('connecting to {0}'.format(client.baseurl))
+
+conference = config.get_conference(client)
+
+conference.close_submissions(freeze_submissions = True)
+
```
pmandler: You should also add the conference to active_venues if it's not already there.

melisabok: it is in this PR: https://github.com/iesl/openreview-py/pull/198

melisabok:
This is already in live, we can merge it.

PR #916, opened 2018-12-20
webfield to release iclr decisions

melisabok:


venues/ICLR.cc/2019/Conference/webfield/conferenceWebfield_decision_tabs.js
```diff
@@ -0,0 +1,147 @@
...
+      } else if (d.content.recommendation === 'Reject') {
+        rejectDecisions.push(notesDict[d.forum]);
+      }
+    }
+  });
+
+  var paperDisplayOptions = {
+    pdfLink: true,
+    replyCount: true,
+    showContents: true
```
zbialecki: I would add showTags: false to the defaults and remove it below.

melisabok:
I'm doing a getAll, it shouldn't have pagination. 

1419 submissions, 1419 meta reviews. and some withdrawn papers.

zbialecki:
Right, OK. I was thinking about how searchResults does pagination, but if you don't pass it a pageSize option then it will render all the notes.

venues/ICLR.cc/2019/Conference/webfield/conferenceWebfield_decision_tabs.js
```diff
@@ -0,0 +1,152 @@
...
+    withdrawnNotes,
+    _.assign({}, paperDisplayOptions, {showTags: false, container: '#withdrawn-papers'})
+  );
+
+
+  $('#notes .spinner-container').remove();
+  $('.tabs-container').show();
+
+  // Show first available tab
+  if (initialPageLoad) {
```
zbialecki: One other thing, this whole `if` block should be removed. I moved this code to the Webfield.ui.done function, and now calling `.click` can cause problems.

melisabok: OK, I will do it

melisabok:
this is in live already and it can be merged.

PR #917, opened 2018-12-21
Change all webfields to use Webfield.ui.done and not call .click 

zbialecki:
Not a critical change, so not all these webfields need to be updated on live. I wanted to change all in case any code was re-used in a future conference. The issue with calling .click on the tab is that doing so will update the browser history and overwrite and existing url hash.


melisabok:
Thanks Zach!

PR #925, opened 2019-01-14
add setup script for multiuser match demo

mspector:
This is a script that you can use to set up the multi-user matching demo. It sets up a dummy ICML conference using reviewers and papers from AKBC.

It picks the first five area chairs and assigns them randomly to papers. For each paper, it takes the original note and adds the assigned AC as a reader. When used with https://github.com/iesl/openreview/pull/1287/files, this should result in assigned ACs being able to see their own papers in the matching browser, and all other papers should show themselves as "Anonymous Paper"

### DIFF
--- a/venues/ICML.cc/2019/Conference/python/icml-multiuser-matching-demo.py
+++ b/venues/ICML.cc/2019/Conference/python/icml-multiuser-matching-demo.py
@@ -1,0 +1,117 @@
+import openreview
+import icml
+import random
+
+client = openreview.Client()
+
+# groups
+icml_groups = openreview.tools.build_groups(icml.conference.id)
+for group in icml_groups:
+    try:
+        existing_group = client.get_group(group.id)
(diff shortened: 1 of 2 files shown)


marshall62:
I pulled this branch and tried running this script but hit the error below.  icml.py is missing a definition for
metadata_inv.   Did you commit your changes to that file?

 File "/home/marshall/iesl/or/scripts/venues/ICML.cc/2019/Conference/python/icml-multiuser-matching-demo.py", line 38, in <module>
    posted_metadata_inv = client.post_invitation(icml.metadata_inv)
AttributeError: module 'icml' has no attribute 'metadata_inv'

mspector:
thanks @marshall62, I just added that back in. Let me know if you run into other issues.

melisabok:
I close this PR but the branch if we want to setup the multiuser UI again.

PR #926, opened 2019-01-14
Akbc19/rebuttal stage updates

mohituniyal:
Merged updates done during ICLR


mspector:
@mohituniyal are these already used on the live site? (i.e. should I just go ahead and merge?)

PR #938, opened 2019-01-24
script to get all accepted papers for Lee

melisabok:
Based on 2018 scripts:

- Adapted to use python 3.

TODO:

- Use pagination to get all the notes.
- Use CSV writer instead of xlsxwriter? 

@mohituniyal could you continue this? it doesn't requiere any special user to get the data.


melisabok:
I got a warning when I try to open the file with Excel but may be it is ok and we don't need to use CSV library:

screen shot 2019-01-24 at 9 33 07 pm

PR #944, opened 2019-01-31
Feature/ldd recruit reviewer

pmandler:


venues/ICLR.cc/2019/Workshop/LLD/python/recruit_reviewers.py
```diff
@@ -0,0 +1,95 @@
...
+with open(args.reviewer_file) as f:
+    for row in csv.reader(f):
+        suggested_reviewers.append(row[0])
+
+formatted_emails=[m.lower() for m in suggested_reviewers]
+
+ids = []
+for member in formatted_emails:
+    try:
+        profile = client.get_profile(member)
```
melisabok: There is a tools.get_profile method that returns None instead of throwing an exception.

melisabok:
Sure, I will add as a TODO the personlization in the builder.

venues/ICLR.cc/2019/Workshop/LLD/python/recruit_reviewers.py
pmandler: I've updated the code to use that.

PR #959, opened 2019-02-06
Colt/fix assignment setup

mspector:
please don't just ignore style conventions. they make the code easier to read.

### DIFF
--- a/venues/learningtheory.org/COLT/2019/python/setup-assignment.py
+++ b/venues/learningtheory.org/COLT/2019/python/setup-assignment.py
@@ -278,13 +278,13 @@ if __name__ == '__main__':
                     'order': 11
                 },
                 'scores_names': {
-                    'values': ['bid'],
+                    'values-dropdown': ['bid'],
                     'required': True,
                     'description': 'List of scores names',
                     'order': 12
                 },
                 'scores_weights': {
-                    'values': ['1.0'], # decimal number allowed
+                    'values-regex': '\\d*\\.?\\d*', # decimal number allowed
                     'required': True,
                     'description': 'Comma separated values of scores weights, should follow the same order than scores_names',
                     'order': 13
--- a/venues/learningtheory.org/COLT/2019/python/setup-assignment.py
+++ b/venues/learningtheory.org/COLT/2019/python/setup-assignment.py
@@ -69,7 +69,7 @@ def _build_entries(author_profiles, reviewer_profiles, paper_bid_jsons, scores_b
             'Very Low': -1.0
         }
         try:
-            reviewer_bids = sorted([t for t in [j for j in paper_bid_jsons if j['tcdate']] if profile.id in t['signatures']], key=lambda t: t.get('tcdate',0), reverse=True)
+            reviewer_bids = sorted([t for t in paper_bid_jsons if profile.id in t['signatures']], key=lambda t: t.get('tmdate',0), reverse=True)
         except TypeError as e:
             print(paper_bid_jsons)
             raise e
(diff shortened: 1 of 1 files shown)


venues/learningtheory.org/COLT/2019/python/setup-assignment.py
```diff
@@ -83,12 +80,13 @@ def _build_entries(author_profiles, reviewer_profiles, paper_bid_jsons, scores_b
             'userid': profile.id,
             'scores': reviewer_scores
         }
+
         # Bid can contain a 'Conflict of Interest' selection, so install in the entry if bid is set to that
         if reviewer_bids:
             tag = reviewer_bids[0]['tag']
             if tag == 'Conflict of Interest':
-                print("Conflict of Interest for", profile.id)
-                user_entry['conflicts'] = 1
```
mspector: by setting the conflicts field to 1, some conflicts will be of type int, and others will be of type list. the encoder will technically allow this, but it's really hacky.

venues/learningtheory.org/COLT/2019/python/setup-assignment.py
```diff
@@ -213,95 +211,98 @@ def post_metadata_note(client,
...
-                    "description": "List of scores names",
-                    "order": 12
+                'scores_names': {
+                    'values': ['bid'],
+                    'required': True,
+                    'description': 'List of scores names',
+                    'order': 12
                 },
-                "scores_weights": {
-                    "values-regex": "\\d*\\.?\\d*", # decimal number allowed
```
mspector: no need to require the user to input a weight here.

venues/learningtheory.org/COLT/2019/python/setup-assignment.py
marshall62: Agreed.  I think that setting the conflicts to an empty list [] will also serve as creating a conflict , but it doesn't really make sense that a list containing nothing indicates a conflict.   So I moved to using 1 which felt more positive as a way to indicate a conflict.   How bout in the future
we use ['conflict'] or something that is both the correct data type (a list) and containing something so it's clear that it's not an empty list that python might treat in an if-condition (see below) as False

One worry I've had about setting things to the empty list  is because I have seen people in this project using of a coding technique where non-boolean tests are placed in the if condition such as:
if user_entry['conflicts']:
   print('hi')

The empty list  is treated as False.

A clearer style would be to always use boolean expressions in conditions 
if len(user_entry['conflicts']) > 0:

mspector: If you look below, I have used ['self-declared COI'].

I disagree that it's clearer to check the length of the list. It's conventional Python to do it the way we've been doing it. From [the PEP-8 style guide](https://www.python.org/dev/peps/pep-0008/): 

> 
> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
> 
> Yes: if not seq:
>      if seq:
> 
> No:  if len(seq):
>      if not len(seq):

mspector: > so it's clear that it's not an empty list that python might treat in an if-condition (see below) as False

We *want* the empty list to evaluate to false. If some part of the matching system is interpreting an empty conflict list to mean that there is a conflict, that is a problem.

marshall62: That (empty list indicated conflict) was my impression which is why I was using 1,  but maybe I'm wrong about the interpretation of the empty list for conflicts in the matcher.   I'll check into this.

melisabok: I think setting a self defined value 

```
user_entry['conflicts'] = ['self-declared COI']
```

 for bid conflicts and merging them with the profile conflicts is the correct behavior:

```
user_entry['conflicts'] = conflicts + user_entry.get('conflicts', [])
```

melisabok:
We found a serious bag that was affected to all the conferences that contain bids. 

This script is filtering tags that not contain a tcdate in the document, **I'm not sure why**. 

The root of the problem is that the backend is setting to null the tcdate when the tag is updated.

PR #970, opened 2019-02-08
read review data from PC member and from AnonReviewer

melisabok:
See status from both types of reviewers

screen shot 2019-02-08 at 2 33 50 pm


melisabok:
screen shot 2019-02-08 at 3 21 44 pm
screen shot 2019-02-08 at 3 21 50 pm
screen shot 2019-02-08 at 3 21 57 pm

PR #975, opened 2019-02-12
Feature/midl metareview

pmandler:
This doesn't work.  When logged in as an Area Chair it says I don't have permission to create a note. What did I miss?


melisabok:
Pam, you have a lot of COLT changes here.

pmandler:
Cripes.  I'll try again.

PR #976, opened 2019-02-12
Feature/midl meta

pmandler:
This needs a fix.  When logged in as an AC it says the user doesn't have permission to create a note.


venues/MIDL.io/2019/Conference/python/create-metareview-invitations.py
```diff
@@ -0,0 +1,98 @@
...
+    paper_num = str(paper.number)
+    paperinv = conference.get_id() + '/-/Paper' + paper_num + '/Meta_Review'
+    paperGroup = conference.get_id() + '/Paper' + paper_num
+
+    ## Confidential Review
+    metareview_reply = {
+        'forum': paper.id,
+        'replyto': paper.id,
+        'writers': {'values-regex': paperGroup + '/Area_Chairs'},
+        'signatures': {'values-regex': paperGroup + '/Area_Chairs'},
```
melisabok: don't you need to set the individual group id as a signature?

like ```Area_Chair1```

melisabok: Take a look to this invitation: https://openreview.net/invitations?id=ICLR.cc/2019/Conference/-/Paper1/Meta_Review

melisabok:
I would be very happy if I add the meta review creation to the builder. How urgent do you need this?

pmandler:
I need it ready for tomorrow.  I agree it belongs in the builder.

melisabok:
ok, let's put this on hold and I can try to do it by today.

melisabok:
@pmandler I implemented the meta review invitation in the builder:

https://github.com/iesl/openreview-py/pull/218/commits/e0bcfc809c480acc9ecd0c73b5878f1b195d1f6b

there is only one difference: the confidence values are different. How did you get those? from MIDL program chairs?

pmandler:
The confidence values are the same as the ones for the reviews which I believe were specified by the program chairs.
Given that it has been noticed that there are no meta-reviews, I'm going ahead with this script and use the builder for the next one.

melisabok:
It makes sense, thanks Pam! I will add a parameter to override these values in the builder.

PR #990, opened 2019-02-14
Fix bidWebfield for new toggling radio buttons

zbialecki:
Fixes issue where changing bids didn't update the tab counts.


melisabok:
Cool! do you mind to add this change to the builder?

https://github.com/iesl/openreview-py/blob/master/openreview/conference/templates/bidWebfield.js

AKBC won't bid anymore.

PR #1005, opened 2019-02-25
Fix/builder refactor

melisabok:
Depends on https://github.com/iesl/openreview-py/pull/230

Set submissions public through the builder.


pmandler:
I'm good with MIDL, KGB and VIVO.

PR #1007, opened 2019-02-27
add commenting support, and add more information to notification emails

mspector:
@melisabok let me know if you have any suggestions!

### DIFF
--- a/venues/OpenReview.net/Support/init.py
+++ b/venues/OpenReview.net/Support/init.py
@@ -155,7 +155,7 @@ request_inv = client.post_invitation(openreview.Invitation(**{
         'readers': {
             'values-copied': [
                 'OpenReview.net/Support',
-                '{content["Contact Emails"]}'
+                '{content["Program Chair Emails"]}'
             ]
         },
         'writers': {
@@ -181,7 +181,7 @@ comment_inv = client.post_invitation(openreview.Invitation(**{
         'invitation': 'OpenReview.net/Support/-/Request_Form',
         'readers': {
             'description': 'Select all user groups that should be able to read this comment.',
-            'values-copied': ['OpenReview.net/Support', '{content["Contact Emails"]}']
+            'values': ['OpenReview.net/Support']
         },
         'writers': {
             'values-copied': [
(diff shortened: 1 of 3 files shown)


venues/OpenReview.net/Support/init.py
```diff
@@ -167,3 +167,46 @@
...
+    'signatures': ['OpenReview.net/Support'],
+    'invitees': ['everyone'],
+    'process': 'commentProcess.js',
+    'reply': {
+        'forum': None,
+        'replyto': None,
+        'invitation': 'OpenReview.net/Support/-/Request_Form',
+        'readers': {
+            'description': 'Select all user groups that should be able to read this comment.',
+            'values': ['OpenReview.net/Support']
```
melisabok: With this only value there is not way to reply to the contact emails.

mspector: Good point. I tried to make a quick fix for this, but I don't think we can get the contact email values from the request form in the comment readers field unless we create a group for them.

(in the request form itself, we can just do values-copied from the Contact Emails field)

melisabok: Does it work having one invitation for all the notes? May be we need to create a comment invitation per forum and add the contact emails there.

venues/OpenReview.net/Support/supportProcess.js
```diff
@@ -2,20 +2,40 @@ function(){
...
     var programchairMailPayload = {
-      'groups': note.content['Contact Emails'],
-      'subject': 'Your request for OpenReview service has been received.',
-      'message': 'You recently requested conference management services from OpenReview. A member of our support team will contact you shortly. You can view the request here: ' + baseUrl + '/forum?id=' + note.forum
+      groups: note.content['Contact Emails'],
+      subject: 'Your request for OpenReview service has been received.',
+      message: 'You recently requested conference management services from OpenReview. A member of our support team will contact you shortly. You can view the request here: ' + baseUrl + '/forum?id=' + note.forum
     };
 
+    var commentInvitation = {
```
melisabok: ❤️

PR #1021, opened 2019-03-08
initial files for LLD

pmandler:
Apparently I forgot to check these in weeks ago.

### DIFF
--- a/venues/ICLR.cc/2019/Workshop/LLD/python/init.py
+++ b/venues/ICLR.cc/2019/Workshop/LLD/python/init.py
@@ -1,0 +1,73 @@
+#!/usr/bin/python
+
+import argparse
+import openreview
+from openreview import tools
+from openreview import invitations
+import config
+import datetime
+
+"""
+OPTIONAL SCRIPT ARGUMENTS
+
+	baseurl -  the URL of the OpenReview server to connect to (live site: https://openreview.net)
+ 	username - the email address of the logging in user
+	password - the user's password
+"""
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--username')
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print('connecting to {0}'.format(client.baseurl))
+
+conference = config.get_conference(client)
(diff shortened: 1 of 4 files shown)


venues/ICLR.cc/2019/Workshop/LLD/python/init.py
```diff
@@ -0,0 +1,73 @@
...
+print('connecting to {0}'.format(client.baseurl))
+
+conference = config.get_conference(client)
+conference.open_submissions(due_date = datetime.datetime(2019, 3, 16, 0, 0))
+
+'''
+Create the homepage and add it to the conference group.
+'''
+conf_group = client.get_group(conference.get_id())
+conf_group.add_webfield('../webfield/conferenceWebfield.js')
```
melisabok: Why do you need to overwrite the webfield?

venues/ICLR.cc/2019/Workshop/LLD/python/init.py
```diff
@@ -0,0 +1,73 @@
...
+conf_group = client.post_group(conf_group)
+print("added webfield to", conf_group.id)
+
+
+'''
+Update process function on submission for blind submission
+'''
+submission_inv =client.get_invitation(conference.get_submission_id())
+with open('../process/submissionProcess.js') as f:
```
melisabok: same here

venues/ICLR.cc/2019/Workshop/LLD/python/init.py
```diff
@@ -0,0 +1,73 @@
...
+    submission_inv.process = f.read()
+
+submission_inv.reply['content']['authorids']['description']+=" Please provide real emails; identities will be anonymized."
+submission_inv.reply['content']['authors']['description']+=" Please provide real names; identities will be anonymized."
+submission_inv = client.post_invitation(submission_inv)
+print("posted invitation "+submission_inv.id)
+
+blind_submission_inv = invitations.Submission(
```
melisabok: we can use the builder to create the blind submissions

pmandler:
These files were made in January before blind submissions were handled by the builder.  I can rewrite this to be current, or we can just merge it in the way that it was used at the time.

melisabok:
ok, but I don't see the web field and process functions in the repo.

pmandler:
Good point.  I've added them.

PR #1031, opened 2019-03-12
fix profileMap & exclude program committee members from email list

mohituniyal:


venues/learningtheory.org/COLT/2019/webfield/programCommitteeWebfield_bidding-stage.js
```diff
@@ -407,7 +408,13 @@ var renderStatusTable = function(individualGroupIds, profiles, notes, completedR
     }
 
     selectedRows.forEach(function(row) {
-      var users = _.values(row[4].reviewers);
+      var reviewers = row[4].reviewers;
+      var users = [];
+      _.forEach(_.keys(reviewers), function(key) {
+        if (key.startsWith('r')) {
```
zbialecki: Unfortunately `.startsWith` isn't available in all browsers, so on the client use `.charAt` for single chars or `.indexOf` for sub string matches.

mohituniyal: Thanks @zbialecki. Fixed.

PR #1033, opened 2019-03-14
Uai/get pdfs

mohituniyal:
Also included in the PR is the change to bid deadline.

### DIFF
--- a/venues/auai.org/UAI/2019/Conference/python/get-pdfs.py
+++ b/venues/auai.org/UAI/2019/Conference/python/get-pdfs.py
@@ -1,0 +1,29 @@
+import requests
+import os
+import openreview
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--baseurl', help="base URL")
+parser.add_argument('--username')
+parser.add_argument('--password')
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print ('connecting to {0}'.format(client.baseurl))
+
+submission_invitation = 'auai.org/UAI/2019/Conference/-/Blind_Submission'
+submissions = openreview.tools.iterget_notes(client, invitation=submission_invitation)
+
+if not os.path.exists('../pdfs'):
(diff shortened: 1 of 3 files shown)


venues/auai.org/UAI/2019/Conference/python/get-pdfs.py
```diff
@@ -0,0 +1,29 @@
...
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print ('connecting to {0}'.format(client.baseurl))
+
+submission_invitation = 'auai.org/UAI/2019/Conference/-/Blind_Submission'
+submissions = openreview.tools.iterget_notes(client, invitation=submission_invitation)
+
+if not os.path.exists('../pdfs'):
+    os.makedirs('../pdfs')
+
+for submission in submissions[:20]:
```
mspector: should this be limited to the first 20 pdfs?

pmandler: Why the 20? If correct please add a comment.

venues/auai.org/UAI/2019/Conference/python/get-pdfs.py
```diff
@@ -0,0 +1,29 @@
...
+parser.add_argument('--password')
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print ('connecting to {0}'.format(client.baseurl))
+
+submission_invitation = 'auai.org/UAI/2019/Conference/-/Blind_Submission'
+submissions = openreview.tools.iterget_notes(client, invitation=submission_invitation)
+
+if not os.path.exists('../pdfs'):
```
mspector: doesn't really matter if we're running this script ourselves, but if you're giving it to someone, I think it should try to create a pdfs directory wherever the user is running the script, not one level up. I also think it should be named something more unique, like "uai18-pdfs" or something (just in case they already have a pdfs directory wherever they're running the script).

(but again, doesn't really matter if we're running it ourselves)

PR #1042, opened 2019-03-18
MIDL main page didn't work for guest

pmandler:


venues/MIDL.io/2019/Conference/Abstract/webfield/conferenceWebfield_abstract.js
```diff
@@ -40,17 +40,27 @@ function load() {
 
   var abstractsP = Webfield.getAll('/notes', { invitation: ABSTRACT_SUBMISSION, details: 'replyCount' });
 
-  userGroupsP = Webfield.get('/groups', { member: user.id, web: true, regex: 'MIDL.io/2019/Conference/(Reviewers|Area_Chairs|Program_Chairs)$' })
+  if (!user || _.startsWith(user.id, 'guest_')) {
```
melisabok: Why this change?

venues/MIDL.io/2019/Conference/Abstract/webfield/conferenceWebfield_abstract.js
```diff
@@ -40,17 +40,27 @@ function load() {
...
+    userGroupsP = $.Deferred().resolve([]);
+      userGroupsP = Webfield.get('/groups', { member: user.id, web: true, regex: 'MIDL.io/2019/Conference/(Reviewers|Area_Chairs|Program_Chairs)$' })
```
melisabok: you shouldn't do this call for guest users

venues/MIDL.io/2019/Conference/Abstract/webfield/conferenceWebfield_abstract.js
```diff
@@ -40,17 +40,27 @@ function load() {
...
 
-  authorNotesP = Webfield.api.getSubmissions(FULL_SUBMISSION, {
-    pageSize: 50,
-    'content.authorids': user.profile.id,
-    details: 'noDetails'
-  });
-
+    authorNotesP = $.Deferred().resolve([]);
+  } else {
+    userGroupsP = Webfield.get('/groups', { member: user.id, web: true }).then(function(result) {
```
melisabok: you need to use the regex to show the consoles. I want to show here 'only' the conference consoles, not the abstract consoles.

melisabok:
You need to use this call:

```
userGroupsP = Webfield.get('/groups', { member: user.id, web: true, regex: 'MIDL.io/2019/Conference/(Reviewers|Area_Chairs|Program_Chairs)$' })
  .then(function(result) {
    return _.map(result.groups, function(g) { return g.id; });
  });
```

instead of:

```
    userGroupsP = Webfield.get('/groups', { member: user.id, web: true }).then(function(result) {
      return _.filter(
        _.map(result.groups, function(g) { return g.id; }),
        function(id) { return _.startsWith(id, CONFERENCE); }
      );
    });
```

The last one will match with Abstract console too and we don't want to show them here

PR #1054, opened 2019-03-22
Colt/rebuttal invi job

mohituniyal:


venues/learningtheory.org/COLT/2019/python/create-rebuttal-invitation-job.py
```diff
@@ -0,0 +1,101 @@
+import openreview
+import config
+import json
+from collections import defaultdict
+
+client = openreview.Client()
```
melisabok: I'm not sure if the instance has the environment variables set

PR #1057, opened 2019-03-27
Feature/recommendations setup script

zbialecki:
Simple script to set up recommendations tag invitation and optionally assign reviewers to papers in the submission group (not required for recommendation interface). This script uses MIDL.amsterdam/2018/Conference as the base group but could be modified to work with any groups by changing the constants at the top of the file.


admin/recommendations-page-setup.py
```diff
@@ -0,0 +1,104 @@
+'''
```
melisabok: Can we move it to the MIDL folder?

zbialecki: OK

PR #1079, opened 2019-04-18
Support/revision python process

mohituniyal:
Once a new request is received on the page https://openreview.net/group?id=OpenReview.net/Support 
 , the support team would see a 'Deploy' button on the request forum allowing them to deploy the venue by providing a conference id.
The authors of the request (program chairs) and the support team will also see a 'Revision' button that would allow updating the support request.
If the venue has already been deployed, the revision would change the field values in the actual venue as well.

Q. I am considering deactivating the 'Deploy' invitation once the conference is deployed. What do you think?

Note: This requires python process support to be deployed on the live site.

### DIFF
--- a/venues/OpenReview.net/Support/revisionProcess.py
+++ b/venues/OpenReview.net/Support/revisionProcess.py
@@ -1,6 +1,3 @@
 def process(client, note, invitation):
-    print('client:', client.baseurl)
-    print('note:', note.id)
-    print('invitation:', invitation.id)
     conference = openreview.helpers.get_conference(client, note.forum)
     print('Conference: ', conference.get_id())
--- a/venues/OpenReview.net/Support/revisionProcess.py
+++ b/venues/OpenReview.net/Support/revisionProcess.py
@@ -3,34 +3,4 @@ def process(client, note, invitation):
     print('note:', note.id)
     print('invitation:', invitation.id)
     conference = openreview.helpers.get_conference(client, note.forum)
-    forum = client.get_note(note.forum)
-    if 'conference_id' in forum.content:
-        # This means that we just updated a live venue
-        readers = forum.content['Contact Emails']
-        readers.append('OpenReview.net/Support')
-        comment_note = openreview.Note(
-            invitation = 'OpenReview.net/Support/-/Request' + str(forum.number) + '/Comment',
-            forum = forum.id,
-            replyto = forum.id,
-            readers = readers,
-            writers = ['OpenReview.net/Support'],
-            signatures = ['OpenReview.net/Support'],
-            content = {
-                'title': 'Your venue has been updated',
-                'comment': '''
-Hi Program Chairs,
-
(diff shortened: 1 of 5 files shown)


venues/OpenReview.net/Support/revisionProcess.py
```diff
@@ -0,0 +1,36 @@
...
+If you need to make a change to the venue, please edit the information provided in your request form directly. We will update your venue accordingly.
+
+If you need special features that are not included in your request form, you can create a comment here or contact us at info@openreview.net and we will assist you.
+
+Thanks!
+
+OpenReview Team'''.format(noteId = forum.id)
+            }
+        )
+        client.post_note(comment_note)
```
melisabok: I'm not sure if we need to create a comment for every change they make. They shouldn't notice that a process has to be run to propagate the changes, it should be transparent for them.

venues/OpenReview.net/Support/supportProcess.js
```diff
@@ -30,7 +30,7 @@ function(){
             values: note.content['Contact Emails'].concat(['OpenReview.net/Support'])
         }
       },
-      signatures: ['OpenReview.net/Support']
```
melisabok: Why are we removing the Support group and using the super user? the process function doesn't work with this group?

melisabok:
I agree with Zach, We can solve that in the UI, similar to the comment/review invitations. If there is already a reply to that invitation then we don't show the button. I can create an issue in the openreview repo.

Let's change the conference_id name to venue_id in the deploy invitation, we can change that name in the builder later.

mohituniyal:
> I think it makes sense to deactivate/hide the deploy button after it's used. I can't think of a situation when we would want to change the conference ID once it's been created.
> 
> My only other comment is that we should consider changing `conference_id` to something more general. Andrew brought this up in the meeting also. Another option might be `venue_id` or `base_id`

Deploy invitation marked multiReply=False and `conference_id` changed to `venue_id`

mohituniyal:
Some more changes to the workflow for Request forms were discussed last week. Following was the finalized workflow:

Event 1: A Program Chair creates a request for venue using the request form.
Resulting System State:
-> Program Chairs and the Support team members are notified via email about the new request.
-> Program Chairs are able to edit the request before the Support team deploys it.
-> Support team is also able to edit the request before deploying it.
-> Support team can see a button called Deploy where they need to provide a venue_id.
-> Both parties can make comments on the request.

Event 2: Support team member deploys the venue
Resulting System State:
-> Program Chairs get an email notification mentioning that their venue has been hosted at a given url. The same is posted as a comment on the request.
-> Program Chairs or the support team can not edit the form directly anymore. Both parties can now post revisions using a 'Revision' button that they would see now. (Edit/Delete buttons are not visible anymore).

Event 3: Program Chairs/ Support Team revise the venue parameters
Resulting System State:
-> A revision to the venue parameters takes effect immediately upon the submission of the revision.

PR #1082, opened 2019-04-25
add init.py file that adds consent agreement to the submission form

mspector:
@melisabok @mohituniyal does this look right? I've tested it and it works, but please let me know if there are other conventions I should be following.

### DIFF
--- a/venues/EMNLP/python/init.py
+++ b/venues/EMNLP/python/init.py
@@ -23,14 +23,10 @@ client = openreview.Client(baseurl=args.baseurl, username=args.username, passwor
 print('connecting to {0}'.format(client.baseurl))
 
 emnlp_request_id = 'rylllwkbKN'
-start_date = datetime.datetime(2019, 4, 2, 13, 30)
-due_date = datetime.datetime(2019, 8, 19, 23, 59)
-
-conference = openreview.helpers.get_conference(client, emnlp_request_id)
 
 author_consent_agreement = (
     '**IMPORTANT: PLEASE READ**\n\n'
-    'The workshop is collecting a research dataset containing a set of peer'
+    'The workshop is collecting a research dataset containing a set of peer '
     'reviews for research purposes. This dataset may be released to the '
     'public domain six months after the final accept/reject decisions are made. '
     'Do you agree to include your reviews in this dataset?'
@@ -39,8 +35,8 @@ author_consent_agreement = (
 additional_fields = {
     "consent": {
         "value-radio": [
-                "Yes, include my anonymized reviews in the \"workshop research dataset\".",
-                "No. Do not include my reviews in the \"workshop research dataset\"."
+            "Yes, include my anonymized reviews in the \"workshop research dataset\".",
+            "No. Do not include my reviews in the \"workshop research dataset\"."
         ],
         "default": "Yes, include my anonymized reviews in the \"workshop research dataset\".",
         "description": author_consent_agreement,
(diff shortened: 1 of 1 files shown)


venues/EMNLP/python/init.py
```diff
@@ -0,0 +1,49 @@
...
+                "Yes, include my anonymized reviews in the \"workshop research dataset\".",
+                "No. Do not include my reviews in the \"workshop research dataset\"."
+        ],
+        "default": "Yes, include my anonymized reviews in the \"workshop research dataset\".",
+        "description": author_consent_agreement,
+        "scroll": True
+    }
+}
+
+conference.open_submissions(additional_fields=additional_fields)
```
melisabok: don't we need to set the start and due date from the request form here?

mspector: I didn't have a great understanding of how the builder worked when I wrote this, but I just looked at the code, and you're right: it does need the dates. I'll fix that.

At the time, I was thinking that `open_submissions` would know about the conference's current state after the `get_conference` call. I thought that I would only need to add information that the request form didn't already have.

One thing I did notice: it looks like `get_conference` is calling `open_submissions` itself. It was surprising to me that a function that has "get" in the name could potentially change the state of the conference (it didn't in this case, because the submissions were already open, so it works fine here). Just thought I would mention it, as a person who is new to the builder/conference API. It could cause trouble in the future.

venues/EMNLP/python/init.py
```diff
@@ -23,6 +23,9 @@
 print('connecting to {0}'.format(client.baseurl))
 
 emnlp_request_id = 'rylllwkbKN'
+start_date = datetime.datetime(2019, 4, 2, 13, 30)
```
melisabok: you should read the dates from the request form note.

venues/EMNLP/python/init.py
melisabok: get_conference gives you an instance of the Conference object with the status reflected in the request note. 

To be sure the objects in the database are synced with the request note we need to call to update all the groups and invitations, otherwise we need to do a diff to check if something has changed. 

```open_submissions``` is not a good name, we should rename it in the future. 

If you want to overwrite something in the invitation you need to call ```open_submissions``` with all the parameters, we currently don't have a method to update the additional fields only.

File: venues/EMNLP/python/init.py (hunk 2 above)
Comment (mspector) [reply venues/EMNLP/python/init.py]: When I went to make this change, I realized that I'd need to copy/paste a lot of the logic from the `get_conference` function. Instead I thought it would be better to make a small change to that function. 

See https://github.com/iesl/openreview-py/pull/280 and let me know what you think. I don't mind copy/pasting if you don't like this solution.

File: venues/EMNLP/python/init.py [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+parser.add_argument('--password')
+
+args = parser.parse_args()
+
+client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+print('connecting to {0}'.format(client.baseurl))
+
+emnlp_request_id = 'rylllwkbKN'
+
+author_consent_agreement = (
```
melisabok: We need to add a revision with these fields otherwise we will loose these changes the next time the builder create the conference without this script, for example: through the process function. 

@mohituniyal can help you to add 'Custom Submission Fields' to the revision invitation. 

One aside note: we should try to use the python convention to name the fields like 'custom_submissions_fields', then the UI will render them nicely, having spaces in json keys can be problematic. There is an open issue for Zach.

mohituniyal:
This has been done through https://github.com/iesl/openreview-scripts/pull/1087
Closing this PR.

PR #1085, opened 2019-05-02
MIDL download review information

pmandler:


melisabok:
I guess this is a script that can be run by the pcs, right?

pmandler:
Hadn't thought about that, but yes it could be.  I don't think they've downloaded openreview to run it themselves.

melisabok:
We could offer them to that, the only thing I would change is the ```config.get_conference()```, you only need the id of group.

PR #1090, opened 2019-05-14
deanonymization script for anonymous preprint server

mspector:


venues/OpenReview.net/Anonymous_Preprint/python/deanonymize.py
```diff
@@ -0,0 +1,50 @@
...
+    mask.content.pop('authors')
+    mask.content.pop('authorids')
+    mask.content['_bibtex'] = mask.content['_bibtex'].replace(
+        '{Anonymous}',
+        ', '.join(original.content['authors']))
+
+    first_author = original.content['authors'][0]
+    lastname = first_author.split(' ')[-1].lower()
+
+    mask.content['_bibtex'] = mask.content['_bibtex'].replace(
```
mspector: not sure if this is really a good way of doing this replacement... let me know what you think

melisabok: We need to move the bibtex generation to our backend soon.

PR #1102, opened 2019-06-14
Feature - Stages (Bid, Review, Meta-review, Decision) added to request form

mohituniyal:
Update: This PR requires https://github.com/iesl/openreview-py/pull/329.

The invitations for initializing these stages is created while deploying the venue.
All the stage-invitiations call reviewProcess.py which checks the invitation calling it and acts only on the appropriate stage. This piece of functionality could have been placed in helpers.py as well but I feel we should not be modifying all stages with every call to openreview.helpers.get_conference(). 
Let me know if people think otherwise.

The stage invitation also create blind notes if the conference is double blind.

Also, reviewer and area chair recruitment has been combined into 1 invitation (screenshot attached).

All the content fields have been names using snake-case (except the review form itself, which will be changed once the snake_case var names are getting rendered correctly).

request form

recruitment-1
recruitment-2


venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -5,7 +5,7 @@ def process(client, note, invitation):
     conference = openreview.helpers.get_conference(client, note.forum)
     print(conference.get_id())
     forum = client.get_note(id=note.forum)
-    comment_readers = forum.content['Contact Emails']
+    comment_readers = forum.content['Contact Emails'][:]
```
pmandler: Why the change? Don't they do the same thing?

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,223 @@ def process(client, note, invitation):
...
+
+To DECLINE the invitation, please click on the following link:
+
+{decline_url}
+
+Please answer within 10 days.
+
+If you accept, please make sure that your OpenReview account is updated and lists all the emails you are using.  Visit http://openreview.net/profile after logging in.
+
+If you have any questions, please contact us at info@openreview.net.
```
pmandler: It says "please contact us", but is signed by the PCs.  
"If you have any questions, please contact info@openreview.net."

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Earlier, "comment_readers" was a reference to the list of contact emails. Any changes to "comment_readers" would actually be changes to the forum.content['Contact Emails']. Also, I am doing a client.post_note(forum) some statements below this. As expected, this was making changes to the "Contact Emails" in the forum's content which was never the intention.
Using [:] creates a shallow copy of the list and assigns this copy to comment_readers, which ensures that changes to comment_readers would not affect the forum object.

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,223 @@ def process(client, note, invitation):
...
+        reply = {
+            'forum': forum.id,
+            'replyto': forum.id,
+            'readers' : {
+                'description': 'The users who will be allowed to read the above content.',
+                'values' : readers
+            },
+            'writers' : {
+                'description' : 'The users who will be allowed to edit the above content',
+                'values' : [conference.get_program_chairs_id(), 'OpenReview.net/Support']
```
pmandler: same as readers?

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,223 @@ def process(client, note, invitation):
...
+                'invitation_email_content': {
+                    'value-regex': '[\\S\\s]{1,10000}',
+                    'description': 'Please carefully review the template below before you click submit to send out recruitment emails. Make sure not to remove the parenthesized tokens.',
+                    'order': 5,
+                    'required': True,
+                    'default': recruitment_email_body
+                }
+            }
+        },
+        signatures = [conference.get_program_chairs_id()]
```
pmandler: Shouldn't it be readers so Support can run this too? Same for the invites below.

venues/OpenReview.net/Support/init.py
```diff
@@ -339,87 +339,222 @@
...
 }))
 
-ac_recruitment_content = {
-    'title': {
-        'value': 'Area chair recruitment',
-        'required': True,
-        'order': 1
+bid_stage_content = {
+    'bid_start_date': {
+        'description': 'When does bidding on submissions begin? Please use the format: YYYY/MM/DD HH:MM (e.g. 2019/01/31 23:59)',
```
pmandler: We need timezone information. Or tell them it is in UTC.

venues/OpenReview.net/Support/init.py
```diff
@@ -339,87 +339,222 @@
...
+        'signatures': {
+            'values-regex': '~.*'
+        },
+        'content': meta_review_stage_content
     }
+}))
 
+decision_stage_content = {
+    'decision_start_date': {
+        'description': 'When does the decision be in the system? Please use the following format: YYYY/MM/DD HH:MM(e.g. 2019/01/31 23:59)',
```
pmandler: "When should the decisions be available to post?" The current wording doesn't make sense.

venues/OpenReview.net/Support/revisionProcess.py
```diff
@@ -1,3 +1,91 @@
...
+        decision_start_date = None
+        if forum_note.content.get('decision_start_date', None):
+            try:
+                decision_start_date = datetime.datetime.strptime(forum_note.content.get('decision_start_date', None), '%Y/%m/%d %H:%M')
+            except ValueError:
+                decision_start_date = datetime.datetime.strptime(forum_note.content.get('decision_start_date', None), '%Y/%m/%d')
+
+        decision_options = forum_note.content.get('decision_options', None)
+        if decision_options:
+            decision_options = [s.translate(str.maketrans('', '', ' "\'')) for s in decision_options.split(',')]
```
pmandler: This is confusing to me.  Overall you're making the string into a list, but I don't get what's going on with the translate, the quotes don't match up. '"\''

mohituniyal: Here, I am splitting the input string decision_options at ',' and then from each output of the split, I am removing spaces, and quote chars ( " and ' ). 
The quotes do not have to match up here as the double quote is inside a single quote space and the following single quote is escaped.

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,223 @@ def process(client, note, invitation):
...
     ))
 
+    recruitment_email_subject = '[{Abbreviated_Venue_Name}] Invitation to serve as {invitee_role}'.format(Abbreviated_Venue_Name = conference.get_short_name())
+    recruitment_email_body = '''Dear {name},
+
+You have been nominated by the program chair committee of {Abbreviated_Venue_Name} to serve as {invitee_role}. As a respected researcher in the area, we hope you will accept and help us make {Abbreviated_Venue_Name} a success.
+
+You are also welcome to submit papers, so please also consider submitting to {Abbreviated_Venue_Name}.
+
+We will be using OpenReview.net and a reviewing process that we hope will be engaging and inclusive of the whole community.
```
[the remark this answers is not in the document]
mohituniyal: Good point.

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Changed. Thanks.

venues/OpenReview.net/Support/init.py
mohituniyal: Thanks for pointing this out. I have changed this to "When will the program chairs start posting decisions?". Let me know if you disagree.

venues/OpenReview.net/Support/revisionProcess.py
mohituniyal: Now that I think about it, maybe removing spaces is not a good idea. I should instead use string.strip() for this to shave off the spaces from the ends.

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,223 @@ def process(client, note, invitation):
...
             'forum': forum.id,
-            'replyto': forum.id,
+            'referent': forum.id,
             'readers' : {
                 'description': 'The users who will be allowed to read the above content.',
                 'values' : readers
             }
         },
-        signatures = ['OpenReview.net/Support']
+        signatures = [conference.get_program_chairs_id()]
```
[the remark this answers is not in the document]
mohituniyal: The signatures here dictate the permissions that the process function of revision will run with and we want all process functions to run with PC permissions so that they get access only within their conference space. OpenReview.net/Support can still these processes.

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Readers additionally had the forum's signatures. I was imagining a scenario where the person filling out the request form is not a program chair. In that case, I wanted to limit access to these invitation to them but still allow them to view the responses made by the actual program chairs. Note that this code has undergone some changes.

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Support can still run these invitations with Program Chair permissions.

venues/OpenReview.net/Support/init.py
mohituniyal: Agreed. This is missing from the request form also. We discussed a few ideas regarding this. Placing some text about this in the description was insufficient as once a reply note is submitted, the note does not render the field description. Let me know if you have any ideas.

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,276 @@ def process(client, note, invitation):
...
-            '''.format(noteId = forum.id, conference_id = conference.get_id(), program_chairs_id = conference.get_program_chairs_id())
+            '''.format(baseurl = client.baseurl, noteId = forum.id, conference_id = conference.get_id(), program_chairs_id = conference.get_program_chairs_id())
         }
     )
     client.post_note(comment_note)
 
+    forum.writers = []
+    forum_readers = [conference.get_program_chairs_id(), 'OpenReview.net/Support']
+    forum_readers.extend(forum.signatures)
+    forum.readers = forum_readers
```
melisabok: Isn't better to do something like this:

```
forum.readers = [conference.get_program_chairs_id(), 'OpenReview.net/Support'] + forum.signatures
```

venues/OpenReview.net/Support/deployProcess.py
melisabok: Isn't bettter to this something like this:

```
comment_readers = forum.content['Contact Emails'] + ['OpenReview.net/Support']
```

venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -35,51 +35,276 @@ def process(client, note, invitation):
...
+                'referent': forum.id,
                 'readers' : {
                     'description': 'The users who will be allowed to read the above content.',
                     'values' : readers
                 }
             },
-            signatures = ['OpenReview.net/Support']
+            signatures = [conference.get_program_chairs_id()]
         ))
+        recruitment_invitation = client.post_invitation(openreview.Invitation(
```
melisabok: Can we define the invitation once and parametrize the value-radio array?

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Agreed. Changes have been made in accordance.

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Agreed. Changes have been made in accordance.

venues/OpenReview.net/Support/deployProcess.py
mohituniyal: Changed accordingly.

PR #1103, opened 2019-06-14
updated support form to move instruction to an accordion widget

mohituniyal:
Screen Shot 2019-06-14 at 12 48 26 PM


mspector:
it looks nice. I approved but I haven't tested myself yet (I will soon)

PR #1111, opened 2019-07-30
Add Support as a member of Conference group, Set Program chair console in revisionProcess

mohituniyal:
Enable reviewer reassignments by default for every new venue.

### DIFF
--- a/venues/OpenReview.net/Support/revisionProcess.py
+++ b/venues/OpenReview.net/Support/revisionProcess.py
@@ -9,6 +9,7 @@ def process(client, note, invitation):
         comment_invitation.reply['readers']['values'] = comment_readers
         updated_comment_invitaiton = client.post_invitation(comment_invitation)
 
+    conference.set_reviewer_reassignment(enabled = True)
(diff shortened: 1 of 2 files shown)


venues/OpenReview.net/Support/revisionProcess.py
```diff
@@ -9,7 +9,8 @@ def process(client, note, invitation):
         conference.set_authors()
         conference.set_reviewers()
         if conference.use_area_chairs:
-            conference.set_area_chairs()
+            conference.set_area_chairs(enable_reviewer_reassignment = True)
```
melisabok: we could create a single property set for the switch on/off of the reassignment.

mohituniyal:
As discussed, we will not update the conference property "enable_reviewer_reassignment" via the Process functions. Instead, it will be set internally based on these factors:
1. Completion of Assignment stage. 
2. If manual matching is requested, we can enable this from the beginning itself for PCs and ACs (beginning would be when blind notes are available, if it is a double blind conference).
Closing this PR.

PR #1124, opened 2019-09-11
Option to add and remove fields from review invitations

mohituniyal:


pmandler:
Can you also add in fields to be removed? This particular venue wants to replace the existing structure with a bunch of different options.

mohituniyal:
> This doesn't actually have an effect on the Official_Reviews. The helpers get_review_stage() needs to handle the new fields, which also means adding the remove fields to the builder for the ReviewStage().

@pmandler  You are right, this change by itself does not affect the Official_Review invitation. It needs the changes in PR https://github.com/iesl/openreview-py/pull/410 to be available first.

PR #1125, opened 2019-09-16
Venue setup - remove option that is not used.

pmandler:
make_reviews_public is not used.  Instead 'Open Reviewing Policy' covers both submissions and reviews.

### DIFF
--- a/venues/OpenReview.net/Support/deployProcess.py
+++ b/venues/OpenReview.net/Support/deployProcess.py
@@ -176,16 +176,6 @@ Program Chairs'''.replace('{Abbreviated_Venue_Name}', conference.get_short_name(
                 'required': True,
                 'order': 11
             },
-            'make_reviews_public': {
-                'description': 'Should the reviews be made public immediately upon posting? Based on your earlier selections, default is "Yes, reviews should be revealed publicly when they are posted".',
-                'value-radio': [
-                    'Yes, reviews should be revealed publicly when they are posted',
-                    'No, reviews should NOT be revealed publicly when they are posted'
-                ],
-                'required': True,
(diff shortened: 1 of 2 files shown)


venues/OpenReview.net/Support/deployProcess.py
```diff
@@ -176,16 +176,6 @@ def process(client, note, invitation):
                 'required': True,
                 'order': 11
             },
-            'make_reviews_public': {
```
melisabok: can we call it 'release_to_public'?

In fact, I think we should merge 'release_reviews_to_authors' too? Like a radio button that says:

```
Release public to:

- Public
- Reviewers and Authors only
- Submitted reviewers only
```

melisabok: The default option is submitted reviewers only I guess

pmandler: For the second choice, it is all reviewers for that paper and authors? I tend to think it should be submitted reviewers and authors.
Also this means we should change the "Open Reviewing Policy" question on setup to be just about submissions, not submissions and reviews.

melisabok: Let's do the mapping of choices with the Review Stage parameters:

- Public (public = True)
- All Paper Reviewers (public = False, release_to_authors = False, release_to_reviewers = True)
- All Paper assigned Reviewers and Authors (public = False, release_to_authors = True, release_to_reviewers = True)
- Only submitted Paper assigned Reviewers (public = False, release_to_authors = False, release_to_reviewers = False) (default option)

melisabok:
I'm going to close this. But we should keep working on releasing the reviews to different readers.

PR #1134, opened 2019-10-03
iclr20 - update review invitations

mohituniyal:
- Setup review invitations
- Add tag invitations to each forum for reviewers to answer the question: "Should this paper have been desk rejected?"

### DIFF
--- a/venues/ICLR.cc/2020/Conference/python/setup-stages.py
+++ b/venues/ICLR.cc/2020/Conference/python/setup-stages.py
@@ -86,7 +86,6 @@ def get_tag_invitation(conference, note, due_date):
         reply = {
             'forum' : note.forum,
             'replyto' : note.forum,
-            "invitation" : conference.get_blind_submission_id(),
             "readers" : {
                 "description": "The users who will be allowed to read the above content.",
                 "values-copied": [
--- a/venues/ICLR.cc/2020/Conference/python/setup-stages.py
+++ b/venues/ICLR.cc/2020/Conference/python/setup-stages.py
@@ -2,6 +2,120 @@ import openreview
 import argparse
 import datetime
 
+remove_review_fields = ['confidence']
+additional_review_fields = {
+    'title': {
+        'order': 1,
+        'value-regex': 'Official Blind Review #[0-9]+',
+        'default': 'Official Blind Review #NUM',
+        'description': 'Please replace NUM with your AnonReviewer number (it is the number following "AnonReviewer" in your signatures below)',
+        'required': True
+    },
+    'review': {
+        'order': 2,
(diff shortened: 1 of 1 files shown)


venues/ICLR.cc/2020/Conference/python/setup-stages.py
```diff
@@ -2,6 +2,120 @@
...
+        'value-radio': [
+            'I have published in this field for several years.',
+            'I have published one or two papers in this area.',
+            'I have read many papers in this area.',
+            'I do not know much about this area.'
+        ],
+        'description': 'Please make a selection that represents your experience correctly',
+        'required': True
+    },
+    'review assessment: thoroughness in paper reading': {
```
melisabok: is it ok to create field name like this one? @zbialecki ?

mohituniyal: Is it the ":" that is troubling you? I can replace it with some other acceptable character. 
PCs wanted three questions in a category, and this was the most straight forward way that I could think of.

zbialecki: I think it should be technically OK, but it might look a little weird in the form.

mohituniyal: This is how it shows up:

Screen Shot 2019-10-03 at 12 56 43 PM

zbialecki: Actually, it looks pretty good to me

melisabok: I'm talking about spaces, we decided to use _ to represent spaces, but I'm not sure about having semicolons in the field names

zbialecki: I see you updated the checkboxes to be radio buttons. It might be better to change the descriptions to something like the description for Experience Assessment.

mohituniyal: Good point @zbialecki. I'll make the change. Thanks.

venues/ICLR.cc/2020/Conference/python/setup-stages.py
```diff
@@ -2,6 +2,119 @@
...
+remove_review_fields = ['confidence']
+additional_review_fields = {
+    'title': {
+        'order': 1,
+        'value-regex': 'Official Blind Review #[0-9]+',
+        'default': 'Official Blind Review #NUM',
+        'description': 'Please replace NUM with your AnonReviewer number (it is the number following "AnonReviewer" in your signatures below)',
+        'required': True
+    },
+    'review': {
```
melisabok: @carlosmondra here please 🙏

PR #1143, opened 2019-11-04
More fixes

melisabok:
this is in live already


melisabok:
Because the job will run in the same instance where the webapp is running and using localhost should be faster

PR #1161, opened 2020-01-31
Rename program chair emails and add contact info

melisabok:


venues/OpenReview.net/Support/init.py
```diff
@@ -44,35 +44,41 @@
...
         'order': 4
     },
-    'Contact Emails': {
+    'program_chair_emails': {
         'description': 'Please provide *lower-cased* email addresses of all the Program Chairs or Organizers (comma-separated) including yourself.',
         'values-regex': '([a-z0-9_\-\.]{2,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,},){0,}([a-z0-9_\-\.]{2,}@[a-z0-9_\-\.]{2,}\.[a-z]{2,})',
         'required': True,
         'order': 5
     },
+    'contact_email': {
```
mohituniyal: Do we want to force venues to have a single point of contact email? If so, I think the description of this field should mention that.
I'm thinking something like "Single point of contact email address which will be displayed in the venue page. Eg. ...".
What do you think @zbialecki @melisabok ?

zbialecki: Yes, I like that wording better. I think it should be "displayed on the venue page" though.

PR #1171, opened 2020-03-12
Fix/post custom load edges

mohituniyal:


venues/thecvf.com/ECCV/2020/Conference/python/post-custom-load-edges.py
```diff
@@ -1,130 +1,129 @@
...
+if __name__ == '__main__':
+    ## Argument handling
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--baseurl', help="base url")
+    parser.add_argument('--username')
+    parser.add_argument('--password')
+    args = parser.parse_args()
+
+    client = openreview.Client(baseurl=args.baseurl, username=args.username, password=args.password)
+    reviewer_group_name = 'Reviewers'
```
melisabok: I'm not sure why this variable exists

venues/thecvf.com/ECCV/2020/Conference/python/post-custom-load-edges.py
```diff
@@ -1,130 +1,129 @@
...
+        else:
+            reduced_loads[note.content['user']] = note
+
+    print ('Reduced loads received: ', len(reduced_loads))
+
+    profile_map = {}
+
+    emails = []
+    tildes = []
+    for ac in reviewer_group.members:
```
melisabok: we should work only with profiles, emails are reviewers with no profile and they will be removed before the paper assignment

venues/thecvf.com/ECCV/2020/Conference/python/post-custom-load-edges.py
mohituniyal: It was added initially when I was unsure if we would need to post custom edges for ACs as well. But now that we know about this I can change this.

venues/thecvf.com/ECCV/2020/Conference/python/post-custom-load-edges.py
mohituniyal: Fixed.

venues/thecvf.com/ECCV/2020/Conference/python/post-custom-load-edges.py
mohituniyal: Fixed.

PR #1176, opened 2020-04-06
Rename options

melisabok:


melisabok:
this is already in the live site, I'm going to merge it.

PR #1177, opened 2020-04-10
script to get paper details required by PCs

mohituniyal:


melisabok:
it seems this script is placed in the ECCV folder?

mohituniyal:
You are right. Sorry about that. I have fixed the path.

PR #1180, opened 2020-04-21
script to get NeurIPS reviewer expertise

mohituniyal:


venues/NeurIPS.cc/2020/Conference/python/get-reviewer-expertise.py
```diff
@@ -0,0 +1,48 @@
...
+                'lname': line[2].strip()}
+        map_profiles = client.search_profiles(emails=list(map_email_to_details.keys()))
+        map_profiles_done = {}
+        csv_writer = csv.writer(f2)
+        csv_writer.writerow(['email', 'firstName', 'lastName', 'profile found', 'active', 'count of publications'])
+        for email, content in tqdm(map_email_to_details.items()):
+            profile = map_profiles.get(email)
+            if profile:
+                if profile.id in map_profiles_done:
+                    continue
```
melisabok, line 43: I'm seeing a difference of 2 lines between the input and output files, do you know why?

why do we have a `continue` here?

mohituniyal: I am aggregating profiles from emails into this dict `map_profiles_done`. I am guessing the missing 2 rows belong to emails that have other emails belonging to the same profile and therefore are already there in the `map_profiles_done`. There is no strong reason for having the `continue`. I can modify it to :
`if profile.id not in map_profiles_done: map_profiles_done[profile.id] = email`.

melisabok: I'm not sure if we should tell them there are two repeated profiles or show the profiles twice, I want to have in and out consistent.

melisabok:
@mohituniyal I added the input parameter so I can run the script from the instance pointing to your csv file.

venues/NeurIPS.cc/2020/Conference/python/get-reviewer-expertise.py
mohituniyal: Ok, I have made some changes and now in and out should remain consistent. I am now additionally printing out any duplicates observed while writing the output.

PR #1185, opened 2020-05-21
Initial script for posting papers

celestemartinez:
-Changed submission invitation.reply['signatures'] to conf_id
-Changed submission invitation.reply['readers'] to everyone
-Changed submission invitation.reply['writers'] to conf_id and PCs

-Main file does not contain author emails, so I create a dictionary with key (name, title) and value email to match author names to emails from a second file

### DIFF
--- a/venues/L4DC/2020/Conference/python/upload-submissions.py
+++ b/venues/L4DC/2020/Conference/python/upload-submissions.py
@@ -5,8 +5,9 @@ import openreview
 from tqdm import tqdm
 
 parser = argparse.ArgumentParser()
-parser.add_argument('l4dc_papers', help='csv file')
+parser.add_argument('--l4dc_papers', help='csv file')
 parser.add_argument('--pdfs_filepath', help='filepath to directory with pdfs')
+parser.add_argument('--author_details', help='file with author names and emails')
 parser.add_argument('--baseurl', help='base url')
 parser.add_argument('--username')
 parser.add_argument('--password')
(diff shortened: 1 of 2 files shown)


venues/L4DC/2020/Conference/python/upload-submissions.py
```diff
@@ -0,0 +1,83 @@
...
+            invitation=conference_id+'/-/Submission',
+            signatures=[conference_id],
+            writers=[conference_id, conference_id+'/Program_Chairs'],
+            readers=['everyone'],
+            content={
+                'title': row[1],
+                'authors': authors,
+                'authorids': authorids,
+                'abstract': row[3],
+                'venue': 'L4DC 2020',
```
melisabok, line 70: I was going to suggest that! :)

melisabok: Should we put a venueid field too?

melisabok: venueid can be the group id: 'L4DC.org/2020' or 'L4DC.org/2020/Conference'.

We should also add bibtex entry. What do you think?

celestemartinez: We need the paper forum for the bibtex url, so I have to first post the paper before adding the bibtex

melisabok:
I think I can merge this, right?

celestemartinez:
> I think I can merge this, right?

Yes please!

PR #1206, opened 2020-12-09
fixing welcome wall invitation and add tags invitations

melisabok:


venues/virtual/python/setup.py
```diff
@@ -866,21 +866,162 @@ def create_presentation_session(day, start, end):
...
+    }
+))
+
+## Free text tags
+tag_invitation_id = f"{virtual_group_id}/-/Tag"
+client.post_invitation(openreview.Invitation(
+    id=tag_invitation_id,
+    readers=['everyone'],
+    writers=[conference_id],
+    invitees=[conference_id],
```
xkopenreview, line 977: for free text tags, should invitees=['~']?

melisabok: yes!

xkopenreview (approved):

PR #1214, opened 2021-09-09
Add scripts to remind assigned revs/acs about reviewing deadline

celestemartinez:


melisabok (approved):

celestemartinez:
@melisabok Made a couple of changes cause I realized in the email the month should be current month (ie. September for these reminders) but the papers and link to consoles are from previous month's iteration (`aclweb.org/ACL/ARR/2021/August`)

melisabok:
Great! I like that, it should work for any venue and the reminder must be sent always before the 13 of each month. Ideally the 10th day of the month.

PR #1223, opened 2022-01-20
improve desk reject actions

melisabok:
- remove unnecessary code from the desk reject process function
- remove get original note
- move the track map to its own file so it can be imported by other files.

### DIFF
--- a/venues/aclweb.org/ACL/add_SACs_as_readers.py
+++ b/venues/aclweb.org/ACL/add_SACs_as_readers.py
@@ -3,6 +3,7 @@ from re import sub
 import openreview
 from tqdm import tqdm
 import csv
+import tracks
 
 """
 OPTIONAL SCRIPT ARGUMENTS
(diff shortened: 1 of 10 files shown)


celestemartinez:
I added the paper title to the email subject as Rodrigo pointed out during the meeting that authors don't know their paper numbers and using the title will help them identify their papers.

celestemartinez (approved):

melisabok:
Cool! you mean the SAC don't know their paper numbers...

celestemartinez:
I meant authors! I see now that we are not sending an email to authors but PCs want us to, so authors know if their paper is desk-rejected

melisabok:
More updates:

- refactor to build the SAC track id
- add the conflict group as nonreaders of all the notes
- fix signature for comments
- use tqdm to check the progress

celestemartinez:
Added:

- a preprocess function for suggested decision invitations which forces SACs to input a ranking if they select 2 or 3 in suggested decision

melisabok:
Thanks for the changes!

PR #1224, opened 2022-01-20
Create rating invitations

celestemartinez:
The invitation ids look like 
`aclweb.org/ACL/2022/Conference/Paper1687/-/MetaReview1907/Rating`
`aclweb.org/ACL/2022/Conference/Paper1687/-/Review6244/Rating`,
which means that all SACs see are 'Rating' buttons in both reviews and metareviews

Note: the PCs wanted the checkbox to be optional and to only include 'Outstanding review/metareview'. It might be a better idea to have a 'not outstanding review/metareview' option and make it required, so that PCs can differentiate between SACs who didn't submit anything and those who don't think the review is outstanding. The form can be submitted without checking anything now which they said was fine but might look weird.

### DIFF
--- a/venues/aclweb.org/ACL/create_rating_inv.py
+++ b/venues/aclweb.org/ACL/create_rating_inv.py
@@ -1,0 +1,122 @@
+import argparse
+import openreview
+from tqdm import tqdm
+import csv
+from openreview import tools
+import re
+from collections import defaultdict
+
+"""
+OPTIONAL SCRIPT ARGUMENTS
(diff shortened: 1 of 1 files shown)


melisabok:
That's super weird, what is the field is called "Outstanding Review" and the values "Yes" or "No" and have No selected by default?

celestemartinez:
I can ask them again but I did offer something like this and they said no so I am not sure

melisabok:
I'm not sure if a note can have an empty content....

melisabok (approved):

PR #1234, opened 2022-02-08
an example pre-check invitation

nlbahy:
This creates a precheck invitation, which could then be added to the 'edit' field of the edge browser to allow ACs to affirm that papers look good and are ready for deployment.


melisabok:
Looks good!

Be aware the button text will say "Pre-check" and we treat the invitation names as noun, I think this is noun and verb?

PR #1243, opened 2022-03-07
Naacl migration script changes

nlbahy:


venues/aclweb.org/NAACL/2022/migrate_submissions_ARR_ACL.py
```diff
@@ -27,11 +27,16 @@
 track_SAC_profiles = {}
 track_groups = { group.id: group for group in client.get_groups('aclweb.org/NAACL/2022/Conference/.*/Senior_Area_Chairs')}
 profile_ids = []
+# Create dictionary for SAC groups and members
+print('Load SAC groups')
```
melisabok, line 31: I think this code is repeated

PR #1246, opened 2022-03-09
finalized naacl invitations

nlbahy:


venues/aclweb.org/NAACL/2022/post_suggested_decision_invitations.py
```diff
@@ -0,0 +1,124 @@
...
+    suggested_decision_super = openreview.Invitation(
+        id = "aclweb.org/NAACL/2022/Conference/-/Suggested_Decision",
+        readers = ["everyone"],
+        writers = ["aclweb.org/NAACL/2022/Conference"],
+        signatures = ["aclweb.org/NAACL/2022/Conference"],
+        reply = {
+            "readers":{
+                "values-copied": [
+                    "aclweb.org/NAACL/2022/Conference",
+                    "aclweb.org/NAACL/2022/Conference/Senior_Area_Chairs"
```
celestemartinez, line 35: missing a comma here at the end, but I guess it doesn't really matter. I think we don't usually set readers, writers and signatures in the super invitation, since we customize these fields for every paper when we submit paper invitations.

celestemartinez (approved):

PR #1260, opened 2023-05-01
Feat: Neurips submission pre-process

celestemartinez:
The pre-process checks two things:
- that the user entered in fields `corresponding_author`, `financial_support` and `reviewer_nomination` is a co-author of the paper
- that if any co-author is a reviewer or AC, they have completed the registration task

Comment: the commented out code is how I tried to do it at first, but it was not working. I *think* it's because the thrown error would exit the inside function, but the outside one didn't know which error was being thrown (the user would just see "Something went wrong")
Is there a better way to do this?

### DIFF
--- a/venues/NeurIPS.cc/2023/submission_preprocess.js
+++ b/venues/NeurIPS.cc/2023/submission_preprocess.js
@@ -44,10 +44,7 @@ async function process(client, edit, invitation) {
     for (const profile of profiles) {
       const emails = profile.content.emails
       const usernames = profile.content.names.map(name => name.username).filter(username => username);
-      const cleanUsernames = usernames.filter(function( element ) {
-        return element !== undefined;
-      });
-      const allIds = emails.concat(cleanUsernames)
+      const allIds = emails.concat(usernames)
       for (const username of allIds) {
         if (reviewerMembers[username]) {
           const { count: noteCount } = await client.getNotes({ invitation: 'NeurIPS.cc/2023/Conference/Reviewers/-/Registration', signatures: allIds })
--- a/venues/NeurIPS.cc/2023/submission_preprocess.js
+++ b/venues/NeurIPS.cc/2023/submission_preprocess.js
@@ -49,33 +49,10 @@ async function process(client, edit, invitation) {
       acMembers[ac] = true
     }
 
-    // var allIds; 
-    // profiles.forEach(async function(profile) {
-    //   const emails = profile.content.emails
-    //   const usernames = profile.content.names.map(name => name.username)
-    //   allIds = emails.concat(usernames)
-    //   for (const username of allIds) {
-    //     if (reviewerMembers[username]) {
-    //       const { count: noteCount } = await client.getNotes({ invitation: 'NeurIPS.cc/2023/Conference/Reviewers/-/Registration', signatures: [allIds] })
-    //       if (noteCount === 0) {
-    //         return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Reviewer ' + username + ' has not completed the Reviewer Registration.' }))
-    //       }
-    //     }
-    //     if (acMembers[username]) {
-    //       const { count: noteCount } = await client.getNotes({ invitation: 'NeurIPS.cc/2023/Conference/Area_Chairs/-/Registration', signatures: [allIds] })
-    //       if (noteCount === 0) {
-    //         return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Area Chair ' + username + ' has not completed the Area Chair Registration.' }))
-    //       }
-    //     }
-    //   }
-    // })
-
-    var allIds; 
-    for(var i=0;i<profiles.length;i++) {
-      const profile = profiles[i]
+    for (const profile of profiles) {
       const emails = profile.content.emails
       const usernames = profile.content.names.map(name => name.username)
-      allIds = emails.concat(usernames)
+      const allIds = emails.concat(usernames)
       for (const username of allIds) {
         if (reviewerMembers[username]) {
           const { count: noteCount } = await client.getNotes({ invitation: 'NeurIPS.cc/2023/Conference/Reviewers/-/Registration', signatures: [allIds] })
--- a/venues/NeurIPS.cc/2023/submission_preprocess.js
+++ b/venues/NeurIPS.cc/2023/submission_preprocess.js
@@ -41,27 +41,26 @@ async function process(client, edit, invitation) {
       reviewerMembers[reviewer] = true
     }
 
-    const { groups: acGroups } = await client.getGroups({ id: invitation.domain + '/Area_Chairs' })
-    const acs = acGroups[0].members;
-    
-    const acMembers = {}
-    for (const ac of acs) {
-      acMembers[ac] = true
-    }
-
     for (const profile of profiles) {
       const emails = profile.content.emails
       const usernames = profile.content.names.map(name => name.username)
       const allIds = emails.concat(usernames)
       for (const username of allIds) {
         if (reviewerMembers[username]) {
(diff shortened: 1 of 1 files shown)


celestemartinez:
Screen Shot 2023-05-01 at 2 11 51 PM

carlosmondra (approved):

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const primaryAuthorField = note.content.corresponding_author.value
+    const primaryAuthorProfile = await client.tools.getProfiles([primaryAuthorField])
+    const primaryAuthor = primaryAuthorProfile ? primaryAuthorProfile[0].id : primaryAuthorField
+    if (!profileIds.includes(primaryAuthor)) {
+      return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Select a paper co-author as corresponding author, ' + primaryAuthorField + ' does not appear in the author list' }))
+    }
+
+    if (note.content.financial_support) {
+      const financialSupportField = note.content.financial_support.value
```
melisabok, line 21: we need to be careful here because the `financial_support` can exist but the value is empty. 

This happens when they put a value and then they edit it clearing the field, then value is an empty string.

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,71 @@
...
+    const { groups: acGroups } = await client.getGroups({ id: invitation.domain + '/Area_Chairs' })
+    const acs = acGroups[0].members;
+    
+    const acMembers = {}
+    for (const ac of acs) {
+      acMembers[ac] = true
+    }
+
+    for (const profile of profiles) {
+      const emails = profile.content.emails
```
melisabok, line 53: don't we need to use confirmedEmails or using emails has more coverage?

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,71 @@
...
+      const usernames = profile.content.names.map(name => name.username)
```
melisabok, line 54: be aware not all the names have a username, I would filter out the usernames that are undefined

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,71 @@
...
+      const allIds = emails.concat(usernames)
+      for (const username of allIds) {
+        if (reviewerMembers[username]) {
+          const { count: noteCount } = await client.getNotes({ invitation: 'NeurIPS.cc/2023/Conference/Reviewers/-/Registration', signatures: [allIds] })
```
melisabok, line 58: isn't `allIds` already an array? I think we need to transform it to a comma separated string array?

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,71 @@
...
+          if (noteCount === 0) {
+            return Promise.reject(new OpenReviewError({ name: 'Error', message: 'Reviewer ' + username + ' has not completed the Reviewer Registration.' }))
```
melisabok, line 60: I think the tools function has a prettyId function that you can use to pretty print the username

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const { groups: revGroups } = await client.getGroups({ id: invitation.domain + '/Reviewers' })
+    const reviewers = revGroups[0].members;
+    
+    const reviewerMembers = {}
+    for (const reviewer of reviewers) {
+      reviewerMembers[reviewer] = true
+    }
+
+    const { groups: acGroups } = await client.getGroups({ id: invitation.domain + '/Area_Chairs' })
```
melisabok, line 44: One small optimization would be to get the AC group after the reviewer validation so avoid one extra network call

melisabok:
have you tested this with real data? how long does it take to do the post operation? I'm worried that it can take too long.

venues/NeurIPS.cc/2023/submission_preprocess.js
celestemartinez: I tried doing this in my local and there is no issue if the value is cleared

carlosmondra: Really? That's odd, I though it was going to fail here :
```js
const financialSupport = financialSupportProfile[0].id
```

venues/NeurIPS.cc/2023/submission_preprocess.js
carlosmondra: I think you are right and we should use `emailsConfirmed`. You cannot sign with unconfirmed emails anyway.

venues/NeurIPS.cc/2023/submission_preprocess.js
carlosmondra:
```suggestion
      const usernames = profile.content.names.map(name => name.username).filter(username => username);
```

venues/NeurIPS.cc/2023/submission_preprocess.js
celestemartinez: could it be that the unconfirmed email is a member of the reviewers group, though?

venues/NeurIPS.cc/2023/submission_preprocess.js
carlosmondra: It's not necessary, the API accepts both arrays and comma separated strings

venues/NeurIPS.cc/2023/submission_preprocess.js
carlosmondra: Right, but no notes could be signed with an unconfirmed email.

venues/NeurIPS.cc/2023/submission_preprocess.js
celestemartinez: it does not go into the if, so we never make it to that code. Maybe I am doing something wrong and not understading the issue here

venues/NeurIPS.cc/2023/submission_preprocess.js
celestemartinez: but we are also using the emails to check if a user is in the reviewers group, if they were added with an unconfirmed email, then we won't catch that this user is a reviewer.

venues/NeurIPS.cc/2023/submission_preprocess.js
carlosmondra: what if `note.content.financial_support.value === ''`? Is this even possible?

venues/NeurIPS.cc/2023/submission_preprocess.js
carlosmondra: aaah I see what you mean. You are right.

venues/NeurIPS.cc/2023/submission_preprocess.js
```diff
@@ -0,0 +1,73 @@
...
+    
+    const reviewerMembers = {}
+    for (const reviewer of reviewers) {
+      reviewerMembers[reviewer] = true
+    }
+
+    for (const profile of profiles) {
+      const emails = profile.content.emails
+      const usernames = profile.content.names.map(name => name.username).filter(username => username);
+      const cleanUsernames = usernames.filter(function( element ) {
```
melisabok, line 47: I guess this filter function can be in the above line, right?

```
const usernames = profile.content.names.map(name => name.username).filter(username => username !== undefined);
```

cc @carlosmondra

celestemartinez: ohh my bad, Carlos gave me the suggestion of doing it in one line and I forgot to remove my implementation to get rid of undefined

melisabok (approved):

PR #1261, opened 2023-05-25
Feature/ Enable chat script

zbialecki:
Add script to enable chat for given list of venues.


zbialecki:
I think this is pretty much ready. I've tested it on a number of different venues locally and used it to deploy the chat to [NeurIPS.cc/2023/Conference](https://dev.openreview.net/group?id=NeurIPS.cc/2023/Conference) on dev.

zbialecki:
This is no longer needed since the features are now in openreview-py

PR #1262, opened 2023-11-03
Upgrade ARR Migration for Commitment Venues

haroldrubio:
This PR replaces the API 1 calls with API 2 calls for commitment venues that are made in API 2. This was only made for individual commitment venues as there is no UI support in the React consoles/webpages


haroldrubio:
There might still be a bug where the migrated submissions are being posted with the same venue ID as the commitment venue

racheljsmart (approved):

PR #1268, opened 2025-09-09
Add paper similarity scripts

enrubio:
This PR shares 2 scripts to compute similarity scores:
- For across 2 venues
- For within the same venue

The final CSV contains paper IDs, author lists, titles, abstracts, and the score.


enrubio:
@melisabok About the merge profiles part of the script where I check if 1 ID belongs to multiple profiles - this is only for within the venue. I remember you mentioned that if it's cross-venue we shouldn't merge.

enrubio:
Closing this since it is being added to the Venue class.

PR #1269, opened 2025-10-07
Add script to transfer AC conflicts

enrubio:
This PR shares a script for transferring conflicts for AC pairs/triplets. It can be used for AC groupings of any size. I generalized the naming of pair/triplet by calling them AC circles to avoid confusion in calling them AC groups.


melisabok:
Can this be a documentation page so PCs can use it ?

enrubio:
Closing for now. We will add this script to the docs and possibly later add this code to openreview-py.

